PR #15, opened 2015-09-02
Close #8 and add citation

### LINKED ISSUE #8: Python path more generic
We should update the paths to the python executable to be more generic. Instead of 

#!/usr/bin/python

we should use

#!/bin/env python

jshoyer:
Sorry about the whitespace differences---emacs automatically added newlines to the ends of files that lacked them. I can go back and split out changes to the hashbang lines only if it bothers any one.

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,11 +21,11 @@ Malia, Max and Noah
 ---
 ##<a id="introduction"></a>Introduction to PlantCV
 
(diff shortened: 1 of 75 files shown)


nfahlgren:
Thanks Steen, the extra line breaks don't bother me.

PR #74, opened 2016-09-13
Cluster contour

maliagehan:
fixed bug in cluster contour split image when there are a different number of names and contours.


maliagehan:
sorry meant to merge with dev.

Issue #106, opened 2017-02-08
Add background subtraction method


nfahlgren:
Cesar found a cool function that lets you subtract an image of the background from an image with both background and foreground (i.e. a picture from our LT system without a plant).

`cv2.BackgroundSubtractorMOG`

Could be interesting to add this function in the future.

PR #116, opened 2017-03-11
Gh background subtraction

calizarr:
Creates the function mentioned in #106


jshoyer:
Based on a quick look, this seems reasonable.
Is there some way to preview the doc pages on readthedocs, or do people have to configure that individually if they want to preview?

Personally I think 'subtract_background' or 'subtract_background_mog' would be better function names than 'background_subtraction', but that is just because I like verbs.

One minor cosmetic issue is that you were .gitignoring some unknown 'change_plantcv.sh' script. The relevant line is at the very bottom of the file (without a newline), under your '# From original .gitignore' heading.

nfahlgren:
@jshoyer it does not look like I can have RTD build pull requests, so the user would need to build the docs locally to view them.

calizarr:
I did forget to remove that since I was testing some stuff out. That can easily be fixed in the next pull request and it really doesn't damage anything to be fair.

I am not attached to function name in any way, so if it seems reasonable to change the name that works too.

I think you have to pull the request and build the docs yourself to actually view them unfortunately.

nfahlgren:
If you need to update anything, any new commit you push should update the pull request automatically. I'm fine with the gitignore changes. I don't have a strong preference for the function name. It would be worth thinking about a naming convention at some point though.

Issue #121, opened 2017-03-19
More flexible image output, especially from 'analyze_object' or related functions


jshoyer:
This is a general feature request that should not require much coding
I have already tested some possibilities that work well enough for my purposes, but I'd like feedback before adding too many new arguments to the analyze_object function.

Examples of altered output I have been using:
  - I just write just the 'shapes' images to disk, not the 'mask' images. Using two arguments (instead of the current 'filename' argument) would make it easy to do the opposite as well.
  - I avoid the color red for 'shapes' lines, because it becomes color-blind-unfriendly with images that have a dark background (i.e top-down views of plants in soil). I have been using magenta, but the color could be specified as an argument.
  - I skip drawing the (blue) contour/object boundaries, because those lines get in the way when I browse my (relatively small, cropped) images.
  - I have been using thinner lines for the same reason---line thickness could be an additional optional argument. (I care less about this one.)

A somewhat related idea I have mentioned is providing options for image output that uses semitransparent background (using `cv2.addWeighted`) instead of straight-up white-masked background, but I should probably file a separate issue for that one before discussing it further.

nfahlgren:
@jshoyer a few comments:

> I just write just the 'shapes' images to disk, not the 'mask' images. Using two arguments (instead of the current 'filename' argument) would make it easy to do the opposite as well.

Makes sense to me that you would be able to select what kind of output images you want, as opposed to our all-or-none method now. For the next release I wouldn't want to introduce any breaking changes to the API though...

> I avoid the color red for 'shapes' lines, because it becomes color-blind-unfriendly with images that have a dark background (i.e top-down views of plants in soil). I have been using magenta, but the color could be specified as an argument.

I agree, red and green should be avoided. I suggest changing the hard-coded color for now and doing something more comprehensive later (see below).

> I have been using thinner lines for the same reason---line thickness could be an additional optional argument. (I care less about this one.)

We made the lines thicker recently because of Jupyter Notebook integration. When we use Matplotlib to plot images in Jupyter it makes smaller, downsampled images (by default), and it tends to make the lines disappear if they are too thin.

[...]

PR #122, opened 2017-03-19
Typo fixes in docstrings etc.

jshoyer:
No changes to code, just comments.

You may wish to defer merging the two lines edited in analyze_objects. It may be better to just remove the 'imgname' argument. I did not go ahead and do that, because I do not know how many scripts that might break.
See also issue #121.


nfahlgren:
@jshoyer thanks! I think we should leave imgname as an argument for now, for compatibility with any scripts out there. Since it's not used it probably doesn't show up in anyone's scripts, but easier to leave it for the time being.

PR #155, opened 2017-06-13
Fix documentation for Sobel Filter

AntonSax:
The sobel filter specifies that "scale" is a parameter, when it is not.
Remove the line conerning "scale"

"dx" and "dy" specify that the derivative can be 1 to 3
Change the documentation to specify the derivative can be from 0 to 3
This is as according to the example usage on the sobel filter documentation.


nfahlgren:
Looks great, thanks @AntonSax!

Issue #167, opened 2017-07-05
Bug: Report Size Marker Limits Circle and Ellipse Location


AntonSax:
### Description
I am trying to use [`report_size_marker_area`](http://plantcv.readthedocs.io/en/latest/report_size_marker/) to define (not detect) a circle in a picture. The function prevents the `x_adj` and `y_adj` parameters from being negative. This works for the when the `shape` parameter is set to `rectangle`, but prevents the `circle` and `ellipse` cases from having a focus at any location above or to the left of the center of the picture.

#### Examples

`device, marker_header,marker_data,images=pcv.report_size_marker_area(img, 'circle', device, debug, marker='define', x_adj=0, y_adj=0, w_adj=-450, h_adj=-450, base='white', objcolor='light', thresh_channel='s', thresh=120, filename=False)`
![269_marker_shape](https://user-images.githubusercontent.com/9059836/27875250-22c27772-6178-11e7-94fb-ba26f5790cdd.png)

`device, marker_header,marker_data,images=pcv.report_size_marker_area(img, 'circle', device, debug, marker='define', x_adj=400, y_adj=0, w_adj=-450, h_adj=-450, base='white', objcolor='light', thresh_channel='s', thresh=120, filename=False)`
![266_marker_shape](https://user-images.githubusercontent.com/9059836/27875254-27038ce0-6178-11e7-8623-37a26c36f99a.png)

`device, marker_header,marker_data,images=pcv.report_size_marker_area(img, 'circle', device, debug, marker='define', x_adj=0, y_adj=400, w_adj=-450, h_adj=-450, base='white', objcolor='light', thresh_channel='s', thresh=120, filename=False)`
![272_marker_shape](https://user-images.githubusercontent.com/9059836/27875257-296a43a2-6178-11e7-940d-a6c0d0df8740.png)

`device, marker_header,marker_data,images=pcv.report_size_marker_area(img, 'circle', device, debug, marker='define', x_adj=0, y_adj=0, w_adj=-300, h_adj=-300, base='white', objcolor='light', thresh_channel='s', thresh=120, filename=False)`
![275_marker_shape](https://user-images.githubusercontent.com/9059836/27875262-2aef6c20-6178-11e7-9b8b-a93eb12eeccb.png)

[...]

PR #168, opened 2017-07-06
Prevent report_size_marker_area from always rejecting negative x_adj and y_adj parameters

AntonSax:
#167 
One way to fix this issue.

This doesn't change that the function initially draws the circle and ellipse from the center of the image.
This only allows the function to accept negative values for `x_adj` and `y_adj` when the shape parameter is set to `circle `or `ellipse`.
This way the circle and ellipse can be drawn anywhere on the image instead of just the bottom right quarter.


nfahlgren:
@AntonSax I think this will work, but the part of the code this affects is not covered by unit testing currently. I think there would need to be some checks added to make sure that the center point of the circle, plus the radius doesn't end up extending outside of the image. @maliagehan can you comment?

maliagehan:
We might need to add in out of bounds checks for the circle and ellipse but shouldn't affect the rectangle.

PR #173, opened 2017-08-28
Distance tranform function

jberry47:
### Checklist:

- [y] Relevant tests (and test data) have been added or updated and passed.
- [y] The documentation was added or updated.
- [y] Relevant issues were updated or added.


jberry47:
I'm assuming that this came from the assert in tests.py

nfahlgren:
The error from the test is `E       AttributeError: 'module' object has no attribute 'distance_transform'`. I think it's because the distance_transform function needs to be added to the import list in the `__init__.py` file. In `__init__.py` you would want to add 'distance_transform' to the `__all__` list and also add a line for `from distance_transform import distance_transform`.

If you add a new commit with those changes it will automatically update the pull request and rerun the tests.

jberry47:
@nfahlgren No problem about the timing. I found the html file and yes that can be removed from the pull request. Thanks!

Issue #182, opened 2017-10-03
Pipeline Parallelization


vissermatthijs:
### Description

Hello, 

I am having problems using the Pipeline Parallelization on a image dataset. When I run the script with the following arguments (Just want to test the script on one image):

time \
/home/matthijs/plantcv/plantcv-pipeline.py \
-d /home/matthijs/PycharmProjects/SMR1/src/vision/ML_background/Data_set/Original \
-p /home/matthijs/PycharmProjects/SMR1/src/vision/scripts/feature_extract.py \
-a filename \
-s plant_db \
-f camera_timestamp_id_other

I get this error:

Starting run 2017-10-03_13-57-42

Building job list... 
Job list will include 1 images

took 7.20024108887e-05

Processing images... 
usage: feature_extract.py [-h] -i IMAGE -o OUTDIR -r RESULT [-w WRITEIMG]
                          [-D DEBUG]
feature_extract.py: error: unrecognized arguments: 03_29_45-Pilot_031014-VIS_TV_z3500.png 03_29_45-Pilot_031014-VIS_TV_z3500.png.txt
took 0.479547977448

Processing results... 
took 0.00504398345947

real	0m0.893s
user	0m0.512s
sys	0m0.348s

It seems that the arguments passed to the my feature_extract.py are not correct? 

The settings of the feature_extract.py pipeline:

#Parse command-line argumentss

def options():
    parser = argparse.ArgumentParser(description="Imaging processing with opencv")
    parser.add_argument("-i", "--image", help="Input image file.", required=True)
    parser.add_argument("-o", "--outdir", help="Output directory for image files.", required=True)
    parser.add_argument("-r", "--result", help="result file.", required=True)
    parser.add_argument("-w", "--writeimg", help="write out images.", default=False)
    parser.add_argument("-D", "--debug", help="Turn on debug, prints intermediate images.", default=False)
    args = parser.parse_args()
    return args
Hope that some one can help me out!

nfahlgren:
@vissermatthijs I think there are potentially two issues. From what I can see above I think there is a space in the image filename and the way I was running the commands caused it to treat the second half of the file as another option incorrectly. I just merged a pull request to update `plantcv-pipeline.py` to fix this, so it should now be able to properly use files with escapable characters.

[...]

PR #183, opened 2017-10-03
Fix potential issues analyzing image files with spaces or escapable characters

nfahlgren:
### Description

`plantcv-pipeline.py` was updated to use the `subprocess.call` method without a shell (either `call([...], shell=True)` or `os.system(...)`). This should improve cross-platform compatibility and is intended to fix potential issues arising from running pipelines on files with spaces and other escapable characters.

### Types of changes

Bug fix

### Checklist:


nfahlgren:
This pull request should partially address issue #182

PR #209, opened 2018-01-23
Guard rogue debug print statements when out of debug.

ic:
### Description

Two `print_image` statements were unconditionally generating input
files. This commit guards the statements according to the debugging
facility logic: File generation happens only when the `debug='print'`
parameter is set.

Note for devops:
This issue is quite important to fix. The unconditional generated
files build up over time, and end up filling deployment disk space.
Production systems suffer from crashes and impossible deployments, as
upgrade packages cannot even be sent to the target machine.

* This PR does not close any open issue, as far as I am aware at opening time.
* Changes are deployed in my production systems, working as expected.

### Types of changes

This is a bug fix.

### Checklist:

- [ ] The documentation was added or updated.
- [ ] Relevant issues were updated or added.

### DIFF
--- a/plantcv/analyze_object.py
+++ b/plantcv/analyze_object.py
@@ -235,10 +235,12 @@ def analyze_object(img, imgname, obj, mask, device, debug=None, filename=False):
         out_file = str(filename[0:-4]) + '_shapes.jpg'
         out_file1 = str(filename[0:-4]) + '_mask.jpg'
 
-        print_image(ori_img, out_file)
+        if debug == 'print':
+            print_image(ori_img, out_file)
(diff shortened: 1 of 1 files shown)


ic:
@nfahlgren Thank you for the feedback.

The reason I came up with this change, is that when `debug=None`, I understood from the documentation that nothing is output. This assumption was false: My production system died by disk full, because of these two images accumulating over time.

---

Thank you for explaining how to actually disable the files. However, is there some valid refactoring to perform here, or perhaps documentation update? Or even the introduction of a new `debug='diagnostic'` mode?

To be honest, as these two files are for diagnostic, I consider them as debug output.

PR #210, opened 2018-01-23
Proposal: Remove duplication of dependency list.

ic:
### Description
The list of library dependencies is currently defined in two places:
* `setup.py`
* `requirements.txt`
This duplication can cause trouble in development, deployment, and
measures of automation.

The proposed change is to elect `requirements.txt` as the master
dependency list, and have `setup.py` load up its content as necessary.

* This change does not close any open issue, at the time of opening.
* I have tested the setup script, but only on Darwin. I have access to Linux also, as needed.

### Types of changes

This change is a refactoring, offering reduced code duplication (and hopefully its positive consequences).

### Checklist:

- [ ] The documentation was added or updated.
- [ ] Relevant issues were updated or added.


ic:
@nfahlgren Thank you for the feedback. I am trying to restore context here, why I proposed that approach without modifying the requirements file at the same time.

I see that my version here uses a nicely packaged [OpenCV for Python](https://github.com/skvark/opencv-python), making setups so much easier. There is also a "contrib" version if PlantCV requires them, but our servers work without so far.

Would this PR be relevant if completed with the above change?

nfahlgren:
@ic I had not seen that before. It looks like a nice way to add OpenCV to the requirements directly without having to do a standalone install. I agree that at least for now we do not need the "contrib" components. I could imagine some specific needs in the future that might require alternative OpenCV builds (maybe like CUDA support or something) but for now that is not an issue, and someone can always install OpenCV themselves of course.

Would you like to update this PR to include `opencv-python` in the requirements file and leave `setup.py` as you already have it?

ic:
@nfahlgren Sorry for taking so long on such a small change. As suggested, I have removed the changes to `setup.py` and added `opencv-python` to PyPi installations.

nfahlgren:
@ic no worries! I think I misspoke before, I was suggesting that by having you add `opencv-python` to `requirements.txt` that we could then keep the changes you made to `setup.py`, i.e. we can remove:

```python
try:
    import cv2
except ImportError:
    raise ImportError("ERROR: OpenCV package 'cv2' not found.")
```

No worries either way though, but if you would like to change `setup.py` back to what you had in the last commit feel free.

ic:
@nfahlgren I restored the changes---re-reading the thread I see my misinterpretation.

ic:
@nfahlgren I do not know what is the reason for the coverage fall. Would this be again something relative to Python 2/3 ?

nfahlgren:
Hi @ic. Sorry for the very long path this took, but your suggestion is now fully implemented in PlantCV, all dependencies are listed in `requirements.txt` and OpenCV is installed with the `opencv-python` PyPI package by default. This took awhile because we tended to use `conda` environments and there's no `conda` package for `opencv-python`, so that made things awkward. To get around this we now deploy PlantCV on PyPI and Bioconda, and of course it can always be built from source: https://plantcv.readthedocs.io/en/latest/installation/.

ic:
@nfahlgren Thank you for all the work! Perhaps we can close this PR ?

By the way, I think `conda` can invoke PyPi. Perhaps good to keep `conda` support, as very popular with data people. Quite a few key packages are not part of `conda` yet, and the delegation to `pip` works fine in my experience.

nfahlgren:
Thanks @ic! You are absolutely right about conda, we have an example of a mixed conda and PyPI environment (https://github.com/danforthcenter/plantcv/blob/master/environment.yml) for installing PlantCV dependencies. We were using it to build our test environment until we switched to pure PyPI for testing (just so we can catch dependency issues faster).

PR #225, opened 2018-04-16
Hyper docs

maliagehan:
for old rotate_img function, made lines thicker for analyze_objects


mkdocs.yml
```diff
@@ -84,7 +84,9 @@ pages:
     - 'Resize': resize.md
     - 'RGB to HSV': rgb2hsv.md
     - 'RGB to LAB': rgb2lab.md
-    - 'Rotate Image': rotate.md
+    - 'Rotate Tools':
```
nfahlgren: I propose we switch the order of the items to put the new function above the one we are deprecating. I also propose a slight name change since "New" tends to get dated pretty fast. How about "Rotate" and "Rotate Image" to match the names of the functions.

docs/rotate2.md
```diff
@@ -0,0 +1,44 @@
+## New Rotate
```
nfahlgren: As with the table of contents, I would propose "Rotate" instead of "New Rotate" just because it won't be new for long.

docs/rotate2.md
```diff
@@ -0,0 +1,44 @@
...
+    
+**Input image**
+
+![Screenshot](img/documentation_images/rotate2/34_whitebalance.jpg)
+
+```python
+import plantcv as pcv
+
+# Rotate image
+device, rotate_img = pcv.rotate_img(img, 10, True, device, debug='print')
```
nfahlgren: `pcv.rotate` instead of `pcv.rotate_img`

docs/rotate2.md
```diff
@@ -0,0 +1,44 @@
...
+
+**Image after rotating 10 degrees**
+
+![Screenshot](img/documentation_images/rotate2/10_rotated_img.jpg)
+
+```python
+import plantcv as pcv
+
+# Rotate image
+device, rotate_img = pcv.rotate_img(img, -10, device, False, debug='print')
```
nfahlgren: `pcv.rotate` instead of `pcv.rotate_img`

plantcv/__init__.py
```diff
@@ -8,7 +8,7 @@
            'crop_position_mask', 'get_nir', 'adaptive_threshold', 'otsu_auto_threshold', 'report_size_marker_area',
            'white_balance', 'triangle_auto_threshold', 'acute_vertex', 'scale_features', 'landmark_reference_pt_dist',
            'x_axis_pseudolandmarks', 'y_axis_pseudolandmarks', 'gaussian_blur', 'cluster_contours',
-           'cluster_contour_splitimg', 'rotate_img', 'shift_img', 'output_mask', 'auto_crop',
+           'cluster_contour_splitimg', 'rotate_img', 'rotate_img2','shift_img', 'output_mask', 'auto_crop',
```
nfahlgren: I think these should be the function names, not the file name, so `rotate` instead of `rotate_img2`.

plantcv/__init__.py
```diff
@@ -68,6 +68,7 @@
 from plantcv.gaussian_blur import gaussian_blur
 from plantcv.cluster_contours import cluster_contours
 from plantcv.cluster_contour_splitimg import cluster_contour_splitimg
+from plantcv.rotate_img2 import rotate
```
nfahlgren: What about naming the file `rotate.py` since the original was `rotate_img.py`, then no need for the version 2 in the filename?

PR #245, opened 2018-07-19
Update Travis-CI to install dependencies with conda instead of pip

nfahlgren:
### Description

Pull request #244 failed because of a build issue with Travis CI and Python3. I think the issue might be that we install OpenCV with Anaconda but install all the other Python dependencies with pip, creating some version mismatch. Since we routinely install everything with Anaconda I updated Travis CI to do the same.

### Types of changes

Buggix

### Checklist:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,9 +27,9 @@ before_install:
   - conda info -a
   # These are the packages we need
   - if [[ "$TRAVIS_PYTHON_VERSION" == "2.7" ]]; then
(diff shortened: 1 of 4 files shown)


nfahlgren:
@maliagehan and @aapokor, this pull request just updates how things are installed in the Travis CI automatic builds because of the issue we're having in pull request #244. Along the way I hit another issue with a newer version of matplotlib where we were saving a plot as a JPG but this is no longer allowed. So basically, nothing really changes here, it just fixes a couple bugs with our build system.

Issue #247, opened 2018-07-28
Some images that fail when running plantcv-pipeline.py are skipped


nfahlgren:
### Description

When doing parallel processing with `plantcv-pipeline.py` some images that "fail" are skipped rather than fail to process correctly. The number of skipped images depends on the number of input images and the number of CPUs requested. The program computes a list of equal-sized lists of jobs for each CPU, with the remainder added to the last CPU. However, the program seems to skip the remainder.

### Details

Parallelization is done using `multiprocessing.map_async` with the list of list of jobs. It seems like the simpler solution that avoids this bug altogether is to instead use `multiprocessing.map` on a list of jobs rather than computing batches of jobs. It might even be faster (and definitely less code).

### Completion Criteria

Once completed, `plantcv-pipeline.py` will at least attempt to process all images in a set. The only failed images will be those that actually fail.

PR #248, opened 2018-07-28
Process in parallel with map

nfahlgren:
Restructure parallelization using a simplified job stack and map. See issue #247

### Description

* The `job_builder` function builds a list of jobs instead of a list of list of jobs. This allows map to iterate over all jobs and replace finished jobs in the process queue immediately rather than splitting up jobs and having each CPU handle batch chunks.
* The `exe_multiproc` function now uses `multiprocessing.map` instead of `multiprocessing.map_async`. Map uses blocking, which is useful in this use case.
* The `process_images_multiproc` now just runs subprocess.call directly instead of iterating over a job list first.

### Types of changes

Bug fix

### Checklist:

### DIFF
--- a/plantcv-pipeline.py
+++ b/plantcv-pipeline.py
@@ -621,9 +621,8 @@ def phenofront_parser(args):
 
 # Process images using multiprocessing
 ###########################################
-def process_images_multiproc(jobs):
-    for job in jobs:
-        call(job)
+def process_images_multiproc(job):
+    call(job)
 
 
 # Multiprocessing pool builder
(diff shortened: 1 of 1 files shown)


nfahlgren:
Note: we don't have tests for `plantcv-pipeline.py` yet. I tested the new method on a set of data from Seren (29 images) with 29 and 2 CPUs and both cases analyzed all 29 images correctly. I also used a sample dataset from @aapokor (100 images) and processed it using 15 CPUs and it processed all 100 correctly (5 failed previously) with the correct CPU utilization.

PR #208, opened 2018-01-16
New feature add init class

nfahlgren:
### Description

This is a small update that adds a class called Params to PlantCV. Params is initialized with default values in the namespace `plantcv.params`. Right now params holds two values, `plantcv.params.device` and `plantcv.params.debug` (default values 0 and None, respectively). This class will be used to simplify our API per discussion in issue #202. Right now there should be no real effect on functionality.

### Types of changes

New feature

### Checklist:

- [ ] Relevant tests (and test data) have been added or updated and passed.
- [ ] The documentation was added or updated.
- [x] Relevant issues were updated or added.

Issue #202, opened 2017-11-15
Simplify and standardize the PlantCV API


nfahlgren:
### Description

This is more of a discussion than a specific task, we can define specific tasks based on discussion generated here.

For PlantCV3 I think we can do a couple of related things: simplify the API (reduce the number of inputs required for each function, merge or split some functions, etc.) and standardize how functions are named. I think it's reasonable to approach this with an eye towards maintaining backwards compatibility, at least for v3.

### Details

Some specific ideas, there could be a lot more added to this:

* Create a PlantCV class (let's call it `params` for now) that is automatically instantiated and contains methods that define tunable parameters (e.g. `device` counter, `debug` mode, etc.).
* Using the PlantCV `params` class, PlantCV functions have access to the `params` methods automatically, so these parameters no longer need to be passed directly to each function (in the example above we immediately remove at least two inputs to every function).
* The user would be able to override the default values in the `params` class to change the behavior of PlantCV.
* Add methods to `params` that would allow users to override default values like line and fill colors, thickness, and other plotting-type features. We would need to change these from hard-coded values in functions to variables that access the `params` class.
* Make new functions where necessary to change or replace the behavior of old functions. The old functions could be rebuilt to call the new version with settings that mimic legacy behavior for backwards compatibility. This could be done for issues like #49 
* In the course of updating any functions (where that makes sense), we can develop a standardized naming approach (and argument order) to make it easier to understand what does what.

### Completion Criteria

We can close this once we finalize a plan and create new, specific issues to move this idea forward (or not).

ic:
The changes you propose would be, from my idiosyncratic use cases, very welcomed. Writing pipelines so far has two major issues:
* Boiler-plate: Many parameters must be repeated all along the pipeline. A language like Python allows so many improvements here.
* At times too thin a layer on top of OpenCV: A higher-lever API would make PlantCV even more useful to plant analysts, without too much prior knowledge of OpenCV. A lower API (the existing one?) could still remain for people with the brains and needs to refine further.

The thread mentions the boiler-plate---big win. I would be curious to hear about several practitioners how a "higher-level API" could be really more useful.

nfahlgren:
@ic thanks for the feedback! I have a pull request in the pipeline for adding the `params` class discussed above, so I think we can start moving on the boiler-plate stuff pretty soon.

Your suggestion about the two API levels is interesting. I can see the value in it for different types of users. I'm trying to think of how it might be organized, happy to have any suggestions from you and others.

[...]

Issue #98, opened 2016-12-15
Boundary Line Tool


scallen81:
Typically the boundary line tool is utilized to distinguish structures (of a vertical image) above the top of a pot using a horizontal line. Alternatively, for horizontal images, for example a whole excavated plant, it would be nice to be able to distinguish between above and below ground structures be demarcating the left from the right with a vertical line.

nfahlgren:
Pull request #217 creates this functionality. Need documentation to complete and close issue.

Issue #97, opened 2016-12-15
Image Rotation


scallen81:
It seems that when one tries to rotate an image, let's say 90 degrees from horizontal to vertical, the image rotates but the frame remains the same. Consequently, if your image is rectangular, when rotated from horizontal to vertical the top and bottom of the image gets cut off. It would be great if both the image and the fame are rotated together.

bhive01:
This probably would help: http://www.pyimagesearch.com/2017/01/02/rotate-images-correctly-with-opencv-and-python/

Seems like it is a known issue.

maliagehan:
yep we had kept image dimensions intentionally, but hadn't thought about when people want to adjust the mat. We'll add the option in soon. Thanks all!

maliagehan:
This option has finally been added, see commit 0d943b20897f9d75c9979000720bf0526346adc6 and new rotate function.

PR #249, opened 2018-08-11
Plantcv v3.0.dev2 staging

nfahlgren:
### Description

This pull request introduces (so far) a range of changes. Primarily, @maliagehan, @aapokor and I have exhaustively gone through the entire package code, tests, and documentation to remove the boilerplate function inputs `device` and `debug` to refactor the package to utilize the `Params` class instead. Now a user can set debug mode once in a script by altering the global debug value (default is `None`). See issues #202 and #208

For example, to enable plotting:

```python
pcv.params.debug = "plot"
img, img_path, img_filename = pcv.readimage("image.png")
sat = pcv.rgb2gray_hsv(img=img, channel="s")
```

In the above example, the images `img` and `sat` will be plotted to the screen without explicitly passing a debug argument to the functions. Likewise, in the background the device counter `pcv.params.device` is used instead of passing a device argument to each function.

Additionally, a new `pcv.params.debug_outdir` can be used to specify where debug images are saved when `pcv.params.debug = "print"` is used. Prior to this debug images could only be saved to the current working directory.

Several functions marked for deprecation have been removed. `define_roi` has been replaced with the `roi` subpackage functions. `analyze_bound` has been replaced with `analyze_bound_horizontal` and `analyze_bound_vertical`. `rotate_img` has been replaced with `rotate`. See issues #97, #98, #217, and #227  

`readimage` was updated to read in images in their native format by default (previously all images were converted to RGB by default). A new `mode` option allows the user to override the default and specify whether the image should be coerced if necessary to RGB or grayscale instead. See issue #49

Several other significant changes may be added to this staged release below, prior to merging.

### Types of changes

New features and functionality. This release will break existing functionality, but relatively straightforward refactoring of existing scripts can restore all prior functionality.

### Checklist:


nfahlgren:
In addition to the comments above, this pull request will alter the PlantCV API in several other ways. Primarily, we went through the code and refactored inputs where we could be more specific about the type of data expected. The vocabulary used was img = any type of image, rgb_img = RGB image only, gray_img = grayscale image only, bin_img = binary image only.

We also moved all of the thresholding functions to a new submodule that is imported automatically.

All changes between the previous and current API are documented in the `updating.md` docs file that will be at https://plantcv.readthedocs.io/en/latest/updating/ after merging.

AdityaKadam30:
getting Error

AttributeError                            Traceback (most recent call last)
<ipython-input-18-fe38899d61dd> in <module>
      1 from plantcv import plantcv as pcv
----> 2 pcv.params.debug = "plot"
      3 img, img_path, img_filename = pcv.readimage("image.png")
      4 sat = pcv.rgb2gray_hsv(img=img, channel="s")

AttributeError: module 'plantcv.plantcv' has no attribute 'params'


Plz help me to solve the error

nfahlgren:
Hi @AdityaKadam30, it sounds like you might need to update your version of PlantCV. Could you check the version you have? On the command line could you change directories to your clone of the PlantCV repository and run the following:

```bash
python setup.py --version
```

The current version would print out `3.0.dev2+72.g9eaad77`

The Params class was not full implemented until at least 3.0.dev2.

If you do need to update you can either use the GitHub Desktop application to Fetch and Pull the latest updates or on the command line run `git pull` in the PlantCV directory, then rerun the `python setup.py install` command.

Issue #259, opened 2018-10-18
Update median_blur inputs 


HaleySchuhl:
### Description
Update kernel dimensions input for median_blur function in plantcv

### Details
Currently takes one input for kernel size but can be updated to specify the length and width.

PR #262, opened 2018-10-19
update median_blur function

HaleySchuhl:
### Description
Function uses spicy.ndimage.filters.median_filter. This allows more flexibility, (ksize X ksize) square kernel or (n,m) box kernel if user speficies a tuple. Added ksize input check to the function. Added test for additional if statement in the function that checks ksize type. median_filter function allows even kernel sizes. Updated documentation. Closes issue #259 

### Checklist:

### DIFF
--- a/plantcv/plantcv/median_blur.py
+++ b/plantcv/plantcv/median_blur.py
@@ -25,7 +25,7 @@ def median_blur(gray_img, ksize):
     """
 
     # Make sure ksize is valid
-    if type(ksize) != int and type(ksize) != tuple:
+    if type(ksize) is not int and type(ksize) is not tuple:
         fatal_error("Invalid ksize, must be integer or tuple")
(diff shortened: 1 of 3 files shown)


plantcv/plantcv/median_blur.py
```diff
@@ -1,28 +1,34 @@
...
+
     :param gray_img: numpy.ndarray
-    :param ksize: int
+    :param ksize: int or tuple
     :return img_mblur: numpy.ndarray
     """
 
-    img_mblur = cv2.medianBlur(gray_img, ksize)
+    # Make sure ksize is valid
+    if type(ksize) != int and type(ksize) != tuple:
```
nfahlgren:
```suggestion
    if type(ksize) is not int and type(ksize) is not tuple:
```

Issue #285, opened 2018-12-14
16-bit requirement in fluor_fvfm


dschneiderch:
I'm running through the PSII tutorial with some of my own images and noticed `fluor_fvfm` requires 16-bit images. any particular reason? we only have 8 bit images.
The main thing I see is the histogram binning - but I think we could easily adjust the number of bins based on the bits.

We also don't capture a dark image - does that need to be required? This one I could deal with by creating a synthetic Fdark image but do you expect everyone to have this as part of their system?

dschneiderch:
and I could change the datatype of my images but I'm not sure if there are any implications for this, assuming I keep my bin # to 256

nfahlgren:
Hi @dschneiderch, I tested the code with 8-bit images and it works (after I removed the 16-bit check). I used 256 bins with the 8-bit images instead of 1000 is the only difference. I have a pull request with this change submitted now along with a few other tweaks.

The only thing we use the Fdark image for is as a quality control check to make sure there's not a light leak or something. I think we were originally going to subtract it from Fmin and Fmax but it's essentially a zero-value image so it didn't have much impact. It would be a bit more work to remove it or make it optional, so if you don't mind supplying a `fdark=np.zeros(np.shape(fmin_img), dtype=np.uint8)` as an input that should do the trick for now.

PS: I added you as a collaborator to this GitHub repository. That will give you some additional capabilities GitHub-wise and will let you directly push to this repo instead of having to work in a Fork. Makes things easier but whichever way works best for you.

PR #286, opened 2018-12-14
Update the Fv/Fm function to allow 8- or 16-bit images

nfahlgren:
### Description

The PSII camera in the LemnaTec system at the Danforth Center outputs 16-bit images so the Fv/Fm analysis function checked that the input images were 16-bit. However, the functionality of the module does not rely on the images being 16-bit, so I removed this check so that 8-bit images could be used. I also updated the pseudocolored image plotting code and documentation.

See issue #285 

### Types of changes

Updates existing function to use 8-bit images in addition to 16-bit images.

### Checklist:

### DIFF
--- a/docs/fluor_fvfm.md
+++ b/docs/fluor_fvfm.md
@@ -2,7 +2,7 @@
 
 Extract Fv/Fm data of objects and produce pseudocolored images.
 
-**plantcv.fluor_fvfm**(*fdark, fmin, fmax, mask, filename, bins=1000*)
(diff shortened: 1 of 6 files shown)


dschneiderch:
Should the function check that fdark, fmin, and fmax are the same bit depth? I added some code but left it commented out because I only checked for 8 or 16  bit consistency. You could presumably use 12 bit or other images. I think different bit depths would cause problems, right?
Also, a default bin of 256 might be more universally applicable than 1000. But if it breaks your code then maybe not worth it. I left this at 1000.

PR #302, opened 2019-01-22
add texture function

HaleySchuhl:
### Description

Describe your changes, for example:
* added docs page for new function
* added a unit test and checked coverage
* made a cropped image for the texture test since it's a computationally expensive function
* added the function to the updating doc
* adjusted binder badge in readme to go directly to the table of contents

### Types of changes

* New feature

### Checklist:


plantcv/plantcv/threshold/threshold_methods.py
```diff
@@ -265,6 +265,59 @@ def triangle(gray_img, max_value, object_type="light", xstep=1):
...
+    :param gray_img: numpy.ndarray
+    :param kernel: int
+    :param threshold: int
+    :param offset: int
+    :param texture_method: str
+    :param borders: str
+    :param max_value: int
+    :return bin_img: numpy.ndarray
+    """
+    from skimage.feature import greycomatrix, greycoprops
```
nfahlgren: The two import statements can go at the top of the file with the other imports. From what I have read it doesn't sound like there's much speed benefit to putting the imports inside a function, unless you want to limit the namespace scope (like we do with matplotlib). Otherwise stylistically the Python preference is at the top of the module.

Issue #318, opened 2019-02-05
VIS imaging tutorials are too confusing


AdamDimech:
I am a beginner to PlantCV and am struggling with the published tutorials for VIS imaging. I know that this issue has been raised before, but it does not satisfactorily help a beginner to python and PlantCV. I am looking to analyse images similar to those in the tutorial.

My starting point was to follow the [VIS tutorial](https://plantcv.readthedocs.io/en/latest/vis_tutorial/)., which results in the [following code](https://gist.github.com/AdamDimech/a863cc3b04d76e922db630276366e76a). (All of the pieces were copied and pasted into one file in sequence). **Issue 1:** Why is there no summary file at the end that one can copy and paste from?

Trial-and-error has shown me that this isn't actually Python-ready and there is no mention of this in the relevant tutorial page. I have figured out that this needs to be saved as a .ipynb and then converted to .py via `jupyter nbconvert --to python VIS_tutorial.ipynb` ([Source](https://plantcv.readthedocs.io/en/latest/jupyter/)). **Issue 2**: The tutorial omits the requirement to add `%matplotlib inline` and some other headers as well. Anyway, following this, a [new .ipynb is created](https://gist.github.com/AdamDimech/830d1de93d8a264322310539d2253fc0). This is then converted to a .py ([see Gist for code](https://gist.github.com/AdamDimech/1f1a9fc20c2c15e18f0401c52f8b871c)).

**Issue 3**: Running the command `VIS_tutorial2.py -i original_image.jpg -o ./output-images -r results.txt 
-w -D 'print'` - [as shown in the tutorial](https://plantcv.readthedocs.io/en/latest/vis_tutorial/) - triggers an error:
```
"usage: VIS_tutorial2.py [-h] -i IMAGE [-o OUTDIR] [-r RESULT] [-w WRITEIMG]
                        [-D DEBUG]
VIS_tutorial2.py: error: argument -w/--writeimg: expected one argument" 
```
Okay, fix and run again: `VIS_tutorial2.py -i original_image.jpg -o ./output-images -r results.txt -w test.jpg -D 'print'`. 
This time the error reads: 

```
File "G:\Path\To\PlantCV\VIS_tutorial2.py", line 140, in <module>
    main()
  File "G:\Path\To\PlantCV\VIS_tutorial2.py", line 51, in main
    s_thresh = pcv.threshold.binary(s, 85, 255, 'light')
NameError: name 's' is not defined
```
[...]

PR #322, opened 2019-02-11
Restructure result outputs

### LINKED ISSUE #319: Add links to complete scripts to tutorials
### Description

As suggested in #318, for tutorials that break down PlantCV scripts (e.g. VIS tutorial: https://plantcv.readthedocs.io/en/latest/vis_tutorial/), we could link to the complete script file for anyone who wants it without having to copy and paste each section.

### Details

Let's add a folder `docs/scripts` and put functional tutorial scripts there (unless this breaks something with `mkdocs`, then we can put them elsewhere). At the top of each tutorial where it makes sense to have a script we can have a link to the script file.

### Completion Criteria

- [ ] Create scripts from tutorials
- [ ] Test scripts to make sure they work
- [ ] Adjust scripts and update tutorials as necessary
- [ ] Add links to scripts in tutorials

HaleySchuhl:
### Title: 
Restructure result outputs

### Description

* Rename the list containing the image for analyze_nir_intensity and analyze_color functions to keep things consistent
* Update the updating.md file
* Created an Outputs class 
* Each function that returns a header and data will also store the data to the Outputs class for easier printing of results
* Analysis images also get saved into Outputs class, but nothing happens with these right now. 
* Added functionality to the `pcv.print_results` function 
* Update unit tests to cover the new print_results function code
________________
* Add detail to parser.add_argument in the VIS tutorial.
* Added script files within the docs folder that contains scripts for each tutorial that allow users to copy and paste into a python file.
* Add links to full scripts in tutorials.
* Updated the way measurement results get written out to use the example of pcv.print_results in each tutorial that had been showing an example of how to format results.
* Realized there hadn't been a documentation page for print_results so added one.
________________
* Added a .clear() method to the Outputs class in the cases where the same analyze function might get run twice (e.g. VIS/NIR tutorial)
* Made a documentation page for the Outputs class modeled after the docs for the Params class
* Added a line to the VIS/NIR tutorial to clear the outputs since the same function is ran twice
* Updated the formatting of print_results sections to make sure everything is ordered the same across images
* Made changes to the following functions so that they store lists containing headers and data to match the format of other functions that return some type of data: landmark_reference_pt_hist,
* Updated docs for the functions above to reflect changes

Closes #319 

Also, `pcv.print_results` will address one of the suggestions made in #318 and eliminate the need for code for writing a summary file at the end of (most) pipelines.

### Checklist:


docs/vis_nir_tutorial.md
```diff
@@ -345,21 +344,7 @@ Write co-result data out to a file.
...
-      coresult.write('\t'.join(map(str,row)))
-      coresult.write("\n")
-    coresult.write('\t'.join(map(str,nshape_header)))
-    coresult.write("\n")
-    coresult.write('\t'.join(map(str,nshape_data)))
-    coresult.write("\n")
-    coresult.write('\t'.join(map(str,nir_imgs[0])))
-    coresult.write("\n")
-    coresult.close()
+    pcv.print_result(filename=args.coresult)
```
nfahlgren: I just realized this will be a problem because the VIS measurements will still be in the output measurements. The NIR measurements will overwrite some but not all unless the user reruns all the same analyses, which we usually don't.

I'm thinking we could add a `clear` method that would delete all current measurements. That would be called by `pcv.outputs.clear()`.

plantcv/plantcv/print_results.py
```diff
@@ -1,18 +1,169 @@
 # Print Numerical Data
```
nfahlgren: For `print_results` I think the output order of the data is not guaranteed to be in the same order as the header because dictionaries are unordered. We could ensure the order by using the header list but I think we can get away with the natural dictionary order. For shape output for example:

```python
if 'shapes' in outputs.measurements:
    header = ["HEADER_SHAPES"]
    data = ["SHAPES_DATA"]
    for k, v in outputs.measurements['shapes'].items():
        header.append(k)
        data.append(v)
    result.write('\t'.join(map(str, header)) + '\n')
    result.write('\t'.join(map(str, data)) + '\n')
```

plantcv/plantcv/print_results.py
```diff
@@ -1,18 +1,169 @@
...
+        result.write('\t'.join(map(str, bound_header)) + '\n')
+        result.write('BOUNDARY_H_DATA')
+
+        for k, v in bound_horizontal_dict.items():
+            result.write('\t' + str(v))
+        result.write('\n')
+
+    if 'bound_vertical' in outputs.measurements:
+        bound_header = [
+            'HEADER_BOUNDARY_V',
```
nfahlgren: The measurements here are not programmed into the `plantcv-pipeline.py` script, but I think we need to fix that over there as opposed to here.

plantcv/plantcv/print_results.py
```diff
@@ -1,18 +1,169 @@
...
-    :param header: list
-    :param data: list
     :return:
     """
-    print('\t'.join(map(str, header)))
-    print('\t'.join(map(str, data)))
+    result = open(filename, "a")
+    if 'bound_horizontal' in outputs.measurements:
+        bound_header = [
+            'HEADER_BOUNDARY_H',
```
nfahlgren: `plantcv-pipeline.py` is expecting the header keyword to be `HEADER_BOUNDARY_#` where # is the boundary line position. I think we will need to refactor that in `plantcv-pipeline.py`. I'm thinking we should add the boundary line position as an output "measurement" for these functions. For the horizontal we store it as `y-position`.

docs/psII_tutorial.md
```diff
@@ -15,6 +15,10 @@ of plants with vertical architecture. You can read more about how we validated t
 analysis pipelines in the [PlantCV Paper](http://dx.doi.org/10.1016/j.molp.2015.06.005). 
 However, the pipelines to analyze PSII images are functional and a sample pipeline is outlined below.  
 
+[![Binder](https://mybinder.org/badge_logo.svg)](https://mybinder.org/v2/gh/danforthcenter/plantcv-binder.git/master?filepath=notebooks/nir_tutorial.ipynb) Check out our interactive NIR tutorial!
```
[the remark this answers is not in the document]
HaleySchuhl: Yes!

plantcv/plantcv/print_results.py
HaleySchuhl: Looking at `plantcv-pipeline.py` I'm a little confused what is happening with the `analyze_bound*` data. It seems like it's only storing data from the horizontal function upon first look? Also, should I change the way landmark data is returned to match the header list, data list return of the other functions?

nfahlgren: Yep, we will need to update `plantcv-pipeline.py` to accommodate somewhat different `analyze_bound_horizontal` output and to add processing for `analyze_bound_vertical`. And you are right about changing the output of the landmark functions to match the other `analyze_*` functions.

plantcv/plantcv/print_results.py
```diff
@@ -1,18 +1,108 @@
...
+        data = ['HISTOGRAM_DATA']
+        for k, v in outputs.measurements['bound_vertical'].items():
+            header.append(k)
+            data.append(v)
+        result.write('\t'.join(map(str, header)) + '\n')
+        result.write('\t'.join(map(str, data)) + '\n')
+
+    # Write data from analyze_color if it was stored
+    if 'color_histogram' in outputs.measurements:
+        header = ['HEADER_COLOR_HISTOGRAM']
```
nfahlgren:
```suggestion
        header = ['HEADER_HISTOGRAM']
```

plantcv/plantcv/print_results.py
```diff
@@ -1,18 +1,108 @@
...
+        for k, v in outputs.measurements['shapes'].items():
+            header.append(k)
+            data.append(v)
+        result.write('\t'.join(map(str, header)) + '\n')
+        result.write('\t'.join(map(str, data)) + '\n')
+
+    # Write data from fluor_fvfm if it was stored
+    if 'fvfm' in outputs.measurements:
+        header = ['HEADER_HISTOGRAM']
+        data = ['FLU_DATA']
```
nfahlgren:
```suggestion
        data = ['HISTOGRAM_DATA']
```

PR #328, opened 2019-02-19
Keep largest contour

### LINKED ISSUE #324: "Keep largest object" option in roi_objects
### Description
* As suggested by @dschneiderch in issue #320 we can add a "keep largest object" in the `pcv.roi_objects` function in addition to the existing "partial" and "cutto" options. This way only the largest contour will be kept. 

### Details
 
How would this change help? Instead of possibly losing smaller plant leaves with filters like `pcv.fill` or `pcv.erode` this option to the `pcv.roi_objects` function would allow users to only keep the largest contour in the case where the object of interest is a solid contour. 

### Completion Criteria

How will we know when this is done?

- [ ] Create method 
- [ ] Write test
- [ ] Add onto documentation 
- [ ] Add to updating.md documentation page

HaleySchuhl:
Title: Add 'keep only largest contour' feature to `roi_objects` function 

### Description
* Added a new option to `roi_type` parameter in `roi_objects` function. If a user sets `roi_type='largest'` then only the largest contour that (at least) partially overlaps with the ROI provided will be kept and returned. All child contours (gaps) will be ignored, so this is not recommended for most image types. 
* Closes #324 
-----------------
* Added onto existing unit tests to cover the new portion of code. 
* Also added more detail to VIS tutorial to describe how to read in an image that is RGBA. 
* Added tutorial scripts to the index so that hopefully links will work now. 

### Types of changes
Is this a: 
* New feature to existing function 

### Checklist:


plantcv/plantcv/roi_objects.py
```diff
@@ -82,6 +83,71 @@ def roi_objects(img, roi_type, roi_contour, roi_hierarchy, object_contour, obj_h
...
+
+        # Draw the contours kept after filtering by ROI
+        kept = cv2.cvtColor(w_back, cv2.COLOR_RGB2GRAY)
+        kept_obj = cv2.bitwise_not(kept)
+        kept_cnt, hierarchy = cv2.findContours(kept_obj, cv2.RETR_TREE, cv2.CHAIN_APPROX_NONE)[-2:]
+
+        # Find the index from the largest contour in the list of contours
+        largest_area = 0
+        for c, cnt in enumerate(kept_cnt):
+            area = cv2.contourArea(cnt)
```
dschneiderch: This looks great! I will test it out later today. A future feature that would be helpful could be to create a filter based on contour area. Instead of filling, eroding, dilating, bluring etc to remove noise, you could just drop small contours. So it would make sense to factor this largest area piece into a new, more general function. Just a thought!

nfahlgren: Would `pcv.fill` work? It evaluates the individual contours by area and removes them if they are smaller than the threshold.

dschneiderch: Yes Now that I looked at the code I think it does the same thing I was thinking about. I thought `pcv.fill` would also fill holes but it doesn't appear to do so. In fact, there is a `remove_small_holes` in `misc.py` that doesn't appear to be used anywhere?

HaleySchuhl: What I’ve done before to fill holes is pcv.invert and then pcv.fill.

PR #330, opened 2019-02-20
add pseudocolor options

### LINKED ISSUE #329: Add pseudocolor function options
### Description
Since `pcv.pseudocolor` is a plotting function we can add more options to the user. 

### Details
* Users are already able to able to choose the type of background (provided they supply a mask), choose their preferred colormap, limit the range of interest for pseudocoloring, auto-crop the plot to the object of interest (provided they supply a single or grouped contour), whether or not the axes are plotted.
* Additional options that we could add include the option to remove the color bar and remove the title.
* Make it so that if the axes are turned off then the title is removed too. 
* Provide defaults so those users that don't have a preference don't have to worry about the additional parameters. 
* Other options?? 

### Completion Criteria
- [ ] Discuss and add other options
- [ ] Add the options
- [ ] Write tests for new options 
- [ ] Update documentation (including more images) 
- [ ] Add to the package (pull request)

HaleySchuhl:
### Description
Describe your changes, for example:
* Updated the function to allow users to remove the colorbar and also the function won't plot the title if axes is set to False
* Update unit testing to cover new if statement
* Update documentation
* Closes #329 

### Types of changes
* Enhancement to an existing function 

### Checklist:


plantcv/plantcv/pseudocolor.py
```diff
@@ -108,38 +115,52 @@ def pseudocolor(gray_img, mask=None, cmap=None, background="image", min_value=0,
         # Plot the background first
         pseudo_img1 = plt.imshow(bkg_img, cmap=bkg_cmap)
```
nfahlgren:
```suggestion
        pseudo_img1 = plt.imshow(bkg_img, cmap=bkg_cmap, vmin=min_value, vmax=max_value)
```

nfahlgren: I think we need to apply the same vmin and vmax to the background...but not 100% sure.

PR #335, opened 2019-02-26
line_thickness parameter

HaleySchuhl:
### Description

- Update the button on the PlantCV documentation homepage to go to the table of contents rather than straight to the VIS tutorial
- Add a line_thickness parameter to the params class
- Update the functions that draw lines so that the thickness will be set to params.line_thickness (and set the default to 5)
- Update functions that draw points (filled in circles) so line thickness is the radius of the circle.
- Updated params.md to include line_thickness
- Functions that include these changes: `analyze_object, all ROI functions, analyze_bound_horizontal, analyze_bound_vertical, acute_vertex, x_axis_pseudolandmark, y_axis_pseudolandmark, scale_features,  roi_objects, object_composition`
- Tested each function with different line thickness values 

### Types of changes

- Enhancement 

### Checklist:


plantcv/plantcv/__init__.py
```diff
@@ -14,11 +14,18 @@ class Params:
     :param device: int
     :param debug: str
     :param debug_outdir: str
+    :param line_thickness: numeric
+    :param line_colorL tuple
```
nfahlgren:
```suggestion
    :param line_color: tuple
```

PR #346, opened 2019-03-10
Bug! should be params.debug_outdir to print

dschneiderch:
Finally figured out why I wasn't seeing my dilate debug images. the print statement was using params.debug instead of params.debug_outdir.
Fixed.


dschneiderch:
Made a couple other changes to dilate.py and erode.py that I thought wuld be helpful. Sorry for the duplicate commit. I changed my commit message and assumed git would be smart enough to sort this out. i hope you can deal with it gracefully.

PR #355, opened 2019-03-22
keep 1st generation sub contours when using 'largest' in roi_objects

### LINKED ISSUE #348: "largest" ROI isn't respecting holes
I'm noticing that when I use the "largest" in `roi_objects` I'm getting holes filled in as objects. Is that a results of only keeping the largest? I suppose that would make sense if all other roi are removed and holes are considered ROI.  Is a way to force holes to be respected within the largest roi?
 Here is the mask where the gap between leaves is clear:
image
But here is the resulting object:
image

dschneiderch:
I tested this again on some images and it seems to work for me with various images and allows `image_composition` to work afterwards.

Closes #348


plantcv/plantcv/roi_objects.py
```diff
@@ -79,16 +77,13 @@ def roi_objects(img, roi_type, roi_contour, roi_hierarchy, object_contour, obj_h
...
+        kept_cnt, kept_hierarchy = cv2.findContours(kept_obj, cv2.RETR_TREE, cv2.CHAIN_APPROX_NONE)[-2:]
+        cv2.drawContours(ori_img, kept_cnt, -1, (0, 255, 0), -1, lineType=8, hierarchy=kept_hierarchy)
         cv2.drawContours(ori_img, roi_contour, -1, (255, 0, 0), params.line_thickness, lineType=8,
                          hierarchy=roi_hierarchy)
 
     # Find the largest contour if roi_type is set to 'largest'
     elif roi_type.upper() == 'LARGEST':
-        # Print warning statement about this feature
-        print(
-            "Warning: roi_type='largest' will only return the largest extreme outer contour. All child contours are left behind.")
```
HaleySchuhl: print("Warning: roi_type='largest' will only return the largest extreme outer contour and its child contours. All grandchild contours are left behind.")

dschneiderch:
Happy to add a warning.  Do you have an example though? I’m having trouble picturing the image.
________________________________

HaleySchuhl:
Under perfect plant segmentation there shouldn't really be the case where plant mass is inside a gap but I zoomed in on one mask I have made in the past. Even though it would be a very small portion of plant I think we should warn users that these small contours will be left behind. 

image

nfahlgren:
@HaleySchuhl and @dschneiderch, can you check to see if the changes I made still do what was intended? I mainly did two things: 1) removed code redundancy, and 2) I got rid of the white-background image and just made the mask directly, which simplified some things. I think the white-background image was originally a feature for visualization but I figured it was redundant with the mask. If anyone disagrees or I messed up the new feature let me know and we can revert back to the previous code.

HaleySchuhl:
@nfahlgren I tested the outputs of the most recent version of `pcv.roi_objects` and they are identical to the expected output. Downstream steps seem to act as we expect also. Thanks for cleaning up the function!

dschneiderch:
roi_objects looks good on my end but I'm confused. What happened to `pcv.pseudocolor`?

HaleySchuhl:
@dschneiderch What do you mean in regards to `pcv.pseudocolor`? I can't find any commits on this branch that edit the pseudocolor function.

dschneiderch:
me neither but the function no longer exists.

image

did you rename it?

HaleySchuhl:
@dschneiderch Oh! Yes, we added a visualize sub-package. So the pseudocolor function is now `pcv.visualize.pseudocolor`. The other changes that came with that was `pcv.plot_hist` is now 'pcv.visualize.histogram` and there is a new visualization tool function `pcv.visualize.colorize_masks`. We grouped these together because we intend these functions to be used purely for creating figures, not returning data or anything else that gets used downstream. Sorry for the confusion!

dschneiderch:
is `pcv.visualize` new since 3.2.0 or part of 3.2.0? it would be worth putting in the release notes somewhere

HaleySchuhl:
@dschneiderch It got added after the 3.2.0 release so it will be in the release notes for 3.3.0. That being said, version 3.3.0 ought to be far more stable than both 3.1 and 3.2. There were a lot of changes made to standardize API, and aside from release notes we also have been trying to keep the <https://plantcv.readthedocs.io/en/latest/updating/> page up to date with changes.

dschneiderch:
@HaleySchuhl I'm using this branch for some processing and when I run with debug='print' I get an empty figure window that opens and immediately closes. The debug plots are printed to file though, I just also get a flashing Figure 1 window as it opens and closes. any issues with this?

dschneiderch:
and if i run my script from the cmd I get 
A1-SILK_dark-2018-12-10 20_01_50-Y3_ALon_fmp

even if I use debug='None'

dschneiderch:
I opened #358 for the figure issue because it occurs with master branch too.

PR #357, opened 2019-03-27
Adding basic statistics to analyze color function

### LINKED ISSUE #314: Analyze color - Mean hue and other statistics
### Description

This is somewhere between a feature request and a discussion. I am interested in analysing how color changes over time. Say I have these images how would be the best way to analyse them?

At the moment my idea, and what I have been testing, is to extract the hue channel from the image, after it has been masked, then counting all the pixels that are from the plant tissue, adding them to an array then calculating the circular mean of the array using scipy's stats.circmean(array, high = 180). After this I would do Statistics on these means to find out if there is a significant difference in colour.

The results from this tally to what I can see looking at histograms, so I beleive it to be a reasonable method, but I thought I would try and get the opinion of others on this idea.

I think it would be a nice addition to the analyze color function, if others agree that this is a reasonable approach to measuring colour.

 Perhaps further statisitics such as median, range etc, would also be helpful.

Something similar to this site http://mkweb.bcgsc.ca/color-summarizer/? 

I am happy to take this on as a project, so long as it seems like a reasonable idea. There may be some logic that I am missing so please critique.

JLJ90:
### Description

- Added calculations for basic statistics for hue values of an image. It still needs some work to alighn with pcv standards.  I have opened this pull request as @HaleySchuhl reccomended this approach for collaboration.
- Closes #314 
### Types of changes

Adds additional functionality to analyze color function
### Checklist:

### DIFF
--- a/docs/analyze_color.md
+++ b/docs/analyze_color.md
@@ -4,7 +4,7 @@ Extract color data of objects and produce pseudocolored images, can extract data
 
 **plantcv.analyze_color**(*rgb_img, mask, bins, hist_plot_type=None*)
 
-**returns** color channel histogram headers, color channel histogram data, color statistics header, color statistics data analysis images  
+**returns** color data headers, color data, data analysis images  
 
 - **Parameters:**  
     - rgb_img - RGB image data
@@ -42,7 +42,7 @@ pcv.params.debug = "print"
 
 # Analyze Color
     
-hist_header, hist_data, color_stats_header, color_stats_data, analysis_images = pcv.analyze_color(rgb_img, mask, 256, 'all')
+color_header, color_data, analysis_images = pcv.analyze_color(rgb_img, mask, 256, 'all')
 ```
 
 **Histograms of (R, G, B), (H, S, V), and (L, A, B) color channels**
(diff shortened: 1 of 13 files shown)


HaleySchuhl:
@JLJ90 I think it makes the most sense to add the new color statistics into the existing header and data that was already getting returned, rather than returning separate lists like I originally suggested. Do you agree @nfahlgren ? I've added onto your pull request (updated the function's documentation page, the updating plantcv documentation page, `plantcv-pipeline.py`, and made sure unit testing still fully covers the function).

nfahlgren:
I made a few changes:
* I had to reconfigure how the measurements were being managed to keep it working along with NIR data, since we store everything signal related in the same output table. We will be working on making it easier to add new measurements, but right now it's a bit tricky.
* I found that hue median and circmean can be calculated on filtered numpy arrays directly, so I simplified the calculations.
* I ended up removing the `bins` input for `analyze_color` because it would only have worked on 8-bit images anyhow and I'm assuming most people did not decrease the bin number from 256. 
* We now output the color channel histograms on their own appropriate scales. RGB = 0-255; H = 0-359 degrees; L, S, and V = 0-100%; and A and B = -127 to 128.
* Tests were updated to remove the `bins` argument.
* Documentation is updated to reflect all of the above.

Issue #423, opened 2019-07-29
parallel pipeline for image series


dschneiderch:
New issue for second question in #422   Should have been separate, sorry.

Original relevant post:
> I took a look at the parallel version for flat files and of course I've been labeling my files from lemnabase differently from you. I am using format code {0}-{3}-{1}-{6} which corresponds to plantid-measurement label-timestamp-camera label whereas you have it setup with measurement label and timestamp flipped.
> Anyway to work on making this more flexible?
> Is this going to iterate through every timestamp?
> 
> here is my repo: https://github.com/dschneiderch/testpipeline

@HaleySchuhl  reply:

> For parallelization over a flat file directory you'll need to specify the labeling format. For example
> -a filename -f id_measurementlabel_timestamp_camera. We do aim to make this more flexible and allow users to provide personalized metadata names outside of the "valid metadata" options. We plan to wrap up v3 soon so this likely won't be implemented until PlantCV v4.
> 
> The default is to iterate through every .png file in the directory but there is the match metadata functionality to filter through files. There is a --dates option that is intended to limit the dates of interest but I've run into problems with this functionality before and haven't been able to successfully debug it.

dschneiderch:
@HaleySchuhl  I think I figured out the metadata names but I had to look in the source. is this documented anywhere at plantcv.readthedocs.io?

So I ran: 
```
plantcv-workflow.py \
-d data/psII \
-a filename \
-p scripts/psii.py \
-j output-parallel.json \
-i debug/parallel \
-f plantbarcode-measurementlabel-timestamp-camera \
-l '-' \
-M measurementlabel:SILK_dark \
-t png
```

However, I get an unknown error:

```
$ ./scripts/run_psii.sh 
Starting run 2019-07-29_10-12-54

Building job list... 
Job list will include 80 images

took 0.35505104064941406

Processing images...
Traceback (most recent call last):
  File "C:/Users/dominikschneider/Miniconda3/envs/test-environment/Scripts/plantcv-workflow.py", line 4, in <module>
    __import__('pkg_resources').run_script('plantcv==3.5.0+16.g5b9908c.dirty', 'plantcv-workflow.py')
  File "C:\Users\dominikschneider\Miniconda3\envs\test-environment\lib\site-packages\pkg_resources\__init__.py", line 666, in run_script
    self.require(requires)[0].run_script(script_name, ns)
[...]

Issue #407, opened 2019-06-25
output from fluor_fvfm


dschneiderch:
Is there a reason that `fluor_fvfm` outputs fv instead of fvfm? also, I think it would make more sense to call this function `fluor_YII()` or even `psII_YII` since you can compute YII for various conditions. Would it be worth a subpackage - it's possible that future use cases could call for qI, qE, etc that are specific to PSII measurements?
currently  i was going to contribute a NPQ function in which case I would presumably output `Fm/Fm' - 1` so i wanted to check with you about the general direction of these functions.

nfahlgren:
Nope, we should output the Fv/Fm image rather than Fv. 

I have also been thinking about renaming the function, but I was thinking it should start with `analyze_` to be consistent with (most) of the other functions that output measurements. In the documentation we refer to the data as PSII related, the references to FLU or FLUOR are because that's the name of the camera system in the LemnaTec software. Since there are other types of FLU camera systems that have other purposes it makes sense to make it clearer what kind of data this function uses.

A subpackage could make sense if there are multiple functions with different inputs. Looks like NPQ would need Fm and Fm' so that makes sense as a separate function.

What if we named the functions given the terminology laid out by Baker (2008), Table 1 specifically [1]? I think while Fv/Fm is used to estimate YII maybe it's clearer to name the function after how we measure YII rather than as YII directly?

The only other consideration is that in the Python community function names should be lowercase, though it's only a convention.

1. Baker NR. 2008. Chlorophyll fluorescence: A probe of photosynthesis in vivo. Annual review of plant biology 59:89–113. DOI: 10.1146/annurev.arplant.59.032607.092759.

dschneiderch:
Ok, `fluor_fvfm.py` can be changed easily enough.
I think following Baker is a good idea (also [Murchie and Lawson 2013](https://academic.oup.com/jxb/article/64/13/3983/436509)  is a nice followup with more practical advice). However, I’m not sure how function naming should work. We don’t need separate functions for Fv/Fm, Fv’/Fm’, Fq’/Fm’. Internally I refer to them as fvfm, fvpfmp, etc but that seems clunky. What about `analyze_psiiefficiency()`?

We could use:
```
analyze_psiiefficiency()   - for psiII and friends
analyze_opencenters()   - qL
analyze_quenching()      - qP
analyze_npq()                 - npq
```
We also need Fo' but I'm not sure what to call that. 

Ultimately, I think it might be more intuitive to have
```
[...]

PR #426, opened 2019-08-14
Photosynthesis functions

dschneiderch:
I added functions for computing YII and NPQ from chlorophyll fluorescence images. YII is general enough and accepts an extra argument to identify the YII, e.g. Fv/Fm or YII from an induction curve. There are also some utility functions for masked statistics.

Several place holder functions also exist for qP qL and fo'. I'd like to include these but we need to figure out the best way of passing Fm to any arbitrary snapshot as discussed in #423 

I have tested these functions against the existing custom functions I was using and get the same results. You can test them with data and scripts from https://github.com/dschneiderch/testpipeline

Relevant tests and documentation have **not** been added.

Please review and let me know if this direction is a good one. Also, let me know if we need changes to integrate with the workflow


HaleySchuhl:
#407

dschneiderch:
closing in favor of #790

Issue #422, opened 2019-07-26
invalid json


dschneiderch:
Finally catching up to development. 
I initialized a test project to get pipelines going. I can run a PSII pipeline with one image pair.
I ran
```
scripts/psii.py -i1 "data\psII\A1-SILK_dark-2018-12-10 20_01_50-PSII0_1_0_0.tif" -i2 "data\psII\A1-SILK_dark-2018-12-10 20_01_50-PSII0_2_0_0.tif" -o debug -D print -writeimg True -r results.json
```
which spit out results.json, but when i try to convert it to a csv i get an error:

```
(test-environment) 
Dominik Schneider@CPPC1 MINGW64 ~/Documents/phenomics/testpipeline (master)
$ plantcv-utils.py json2csv -j results.json -c result-table
Traceback (most recent call last):
  File "C:/Users/dominikschneider/Miniconda3/envs/test-environment/Scripts/plantcv-utils.py", line 4, in <module>
    __import__('pkg_resources').run_script('plantcv==3.5.0+16.g5b9908c.dirty', 'plantcv-utils.py')
  File "C:\Users\dominikschneider\Miniconda3\envs\test-environment\lib\site-packages\pkg_resources\__init__.py", line 666, in run_script    
    self.require(requires)[0].run_script(script_name, ns)
  File "C:\Users\dominikschneider\Miniconda3\envs\test-environment\lib\site-packages\pkg_resources\__init__.py", line 1453, in run_script   
    exec(code, namespace, namespace)
  File "c:\users\dominikschneider\miniconda3\envs\test-environment\lib\site-packages\plantcv-3.5.0+16.g5b9908c.dirty-py3.7.egg\EGG-INFO\scripts\plantcv-utils.py", line 74, in <module>
    main()
  File "c:\users\dominikschneider\miniconda3\envs\test-environment\lib\site-packages\plantcv-3.5.0+16.g5b9908c.dirty-py3.7.egg\EGG-INFO\scripts\plantcv-utils.py", line 69, in main
    options()
  File "c:\users\dominikschneider\miniconda3\envs\test-environment\lib\site-packages\plantcv-3.5.0+16.g5b9908c.dirty-py3.7.egg\EGG-INFO\scripts\plantcv-utils.py", line 44, in options
    args.func(args)
[...]

PR #434, opened 2019-08-23
update gitignore and pipeline parallel doc

dschneiderch:
Sorry I got some confusion happening in my branch. I think I"ve resolved my issues though. Here are the differences i had on my computer that I wanted to submit.

Mostly I wanted to add .vscode to the .gitignore 

In pipeline_parallel.md I also had a slightly more explicit doc statement for the image format flag.

In process_results.py I added json to the comment, but I also now see that we are checking text and json. I thought the resolution of #422 was to check for json only? should this be changed?


nfahlgren:
I think we are checking with both for now because `plantcv-workflow.py` is still creating .txt files at the moment.

Issue #453, opened 2019-10-02
no debug output


dschneiderch:
I think I'm still generally confused by the workflow pipelines.

I'm following your [VIS Tutorial](https://plantcv.readthedocs.io/en/stable/vis_tutorial/) with my own image.
https://emailwsu-my.sharepoint.com/:i:/g/personal/dominik_schneider_wsu_edu/EfTahEzq8jpEsXkPB9ZaOiwBDBHwefipD9W1LxtXDhF6pA?e=c09Ty2

I'm running this at the commandline:

```
python vis_workflow.py -i "data\D7-chickpea_test_slowgrowth-2019-09-27 04_16_23-VIS0.png"  -o debug-images -r results.txt -w -D 'print'
```

This should produce a results.txt (True) but also print debug images to debug-images (False).... correct?
Running this command does not produce any errors. I also tried renamiing the image to have no delimiters but it didn't change anything.
Why does the `-w` flag not accept True or False?

Also, I won't be able to run over the entire directory of .png images until the date thing is sorted, correct?

Windows, Conda, plantcv 3.6 via bioconda
 

```
# !/usr/bin/python
import sys, traceback
import cv2
import numpy as np
import argparse
import string
from plantcv import plantcv as pcv
import os
import glob
import numpy as np
from matplotlib import pyplot as plt


### Parse command-line arguments
def options():
    parser = argparse.ArgumentParser(
        description="Imaging processing with opencv")
    parser.add_argument("-i",
                        "--image",
                        help="Input image file.",
                        required=True)
    parser.add_argument("-o",
                        "--outdir",
                        help="Output directory for image files.",
                        required=False)
    parser.add_argument("-r", "--result", help="result file.", required=False)
    parser.add_argument("-w",
                        "--writeimg",
                        help="write out images.",
                        default=False,
                        action="store_true")
    parser.add_argument(
        "-D",
        "--debug",
        help=
        "can be set to 'print' or None (or 'plot' if in jupyter) prints intermediate images.",
        default=None)
    args = parser.parse_args()
    return args


#### Start of the Main/Customizable portion of the workflow.


### Main workflow
def main():
    # Get options
    args = options()

    pcv.params.debug = args.debug  # set debug mode
    pcv.params.debug_outdir = args.outdir  # set output directory

    # Read image
    img, path, filename = pcv.readimage(filename=args.image)

    img_wb = pcv.white_balance(img, mode = 'max', roi = (1455,3085,100,100))
[...]

PR #454, opened 2019-10-07
Add regular expression-based metadata parser option to plantcv.parallel

nfahlgren:
**Describe your changes**
The current metadata parser in `plantcv.parallel` requires image names to be structured with a consistent delimiter such that metadata terms can be split apart using the input delimiter character. Given that data collection occurs on heterogeneous systems with data output in formats users often cannot control, this option does not always work (see #423 and #453). This pull request attempts to address this issue by adding an advanced input option that allows the user to input a regular expression pattern that can be used to parse image names in a flexible manner. Itemized changes:

* Change the expected input format for the `--meta` option to comma-separated values rather than using the `--delimiter` option.

**Type of update**
Is this a new feature.

**Associated issues**
Partially addresses:
#423 
#453


dschneiderch:
This did not work for me. Looks like something broke.


1. 

v3.6 - this works with these filenames
data\slowgrowth\D7-chickpea_test_slowgrowth-20190927 041623-VIS0.png
```
python.exe ^
%CONDA_PREFIX%\Scripts\plantcv-workflow.py ^
--dir data\slowgrowth ^
--workflow vis_workflow.py ^
--type png ^
--json slowgrowth.json ^
--outdir output-images ^
--adaptor filename ^
--delimiter - ^
--meta plantbarcode-measurementlabel-timestamp-camera ^
-T 6 ^
-D 2019-10-05 ^
-w ^
--create ^
--other_args="--debugdir debug-images"
```

2.

this branch of 3.6.1, same filenames, as documented:
```
(test-environment) C:\Users\dominikschneider\Documents\phenomics\chickpea>run_workflows.cmd

python.exe C:\Users\dominikschneider\Miniconda3\envs\test-environment\Scripts\plantcv-workflow.py --dir data\slowgrowth --workflow vis_workflow.py --type png --json slowgrowth.json --outdir output-images --adaptor filename --delimiter '(.{2})-(.+)-(\d{4}\d{2}\d{2} \d{2}\d{2}\d{2})-(.+)' --meta plantbarcode,measurementlabel,timestamp,camera -T 6 -D 2019-10-05 -w --create --other_args="--debugdir debug-images"
Starting run 2019-10-17_15-32-14

usage: plantcv-workflow.py [-h] -d DIR [-a ADAPTOR] -p WORKFLOW -j JSON -f
                           META [-i OUTDIR] [-T CPU] [-c] [-D DATES] [-t TYPE]
                           [-l DELIMITER] [-M MATCH] [-C COPROCESS] [-w]
                           [-o OTHER_ARGS]
plantcv-workflow.py: error: unrecognized arguments: \d{2}\d{2}\d{2})-(.+)'
```

Windows does NOT like spaces or single quotes.

3. 

with double quotes around the regex `"(.{2})-(.+)-(\d{4}\d{2}\d{2} \d{2}\d{2}\d{2})-(.+)"`
```
(test-environment) C:\Users\dominikschneider\Documents\phenomics\chickpea>run_workflows.cmd

python.exe C:\Users\dominikschneider\Miniconda3\envs\test-environment\Scripts\plantcv-workflow.py --dir data\slowgrowth --workflow vis_workflow.py --type png --json slowgrowth.json --outdir output-images --adaptor filename --delimiter "(.{2})-(.+)-(\d{4}\d{2}\d{2} \d{2}\d{2}\d{2})-(.+)" --meta plantbarcode,measurementlabel,timestamp,camera -T 6 -D 2019-10-05 -w --create --other_args="--debugdir debug-images"
Starting run 2019-10-17_15-46-54

Building job list...
Job list will include 4 images

took 0.02692723274230957

Processing images...
Traceback (most recent call last):
  File "C:\Users\dominikschneider\Miniconda3\envs\test-environment\Scripts\plantcv-workflow.py", line 4, in <module>
    __import__('pkg_resources').run_script('plantcv==3.6.1+6.g79605c9', 'plantcv-workflow.py')
  File "C:\Users\dominikschneider\Miniconda3\envs\test-environment\lib\site-packages\pkg_resources\__init__.py", line 666, in run_script
    self.require(requires)[0].run_script(script_name, ns)
  File "C:\Users\dominikschneider\Miniconda3\envs\test-environment\lib\site-packages\pkg_resources\__init__.py", line 1453, in run_script
    exec(code, namespace, namespace)
  File "c:\users\dominikschneider\miniconda3\envs\test-environment\lib\site-packages\plantcv-3.6.1+6.g79605c9-py3.7.egg\EGG-INFO\scripts\plantcv-workflow.py", line 319, in <module>
    main()
  File "c:\users\dominikschneider\miniconda3\envs\test-environment\lib\site-packages\plantcv-3.6.1+6.g79605c9-py3.7.egg\EGG-INFO\scripts\plantcv-workflow.py", line 292, in main
    pcvp.multiprocess(jobs, args.cpu)
  File "c:\users\dominikschneider\miniconda3\envs\test-environment\lib\site-packages\plantcv-3.6.1+6.g79605c9-py3.7.egg\plantcv\parallel\multiprocess.py", line 15, in multiprocess
    p = mp.Pool(processes=cpus)
  File "C:\Users\dominikschneider\Miniconda3\envs\test-environment\lib\multiprocessing\context.py", line 119, in Pool
    context=self.get_context())
  File "C:\Users\dominikschneider\Miniconda3\envs\test-environment\lib\multiprocessing\pool.py", line 176, in __init__
    self._repopulate_pool()
  File "C:\Users\dominikschneider\Miniconda3\envs\test-environment\lib\multiprocessing\pool.py", line 241, in _repopulate_pool
    w.start()
  File "C:\Users\dominikschneider\Miniconda3\envs\test-environment\lib\multiprocessing\process.py", line 112, in start
    self._popen = self._Popen(self)
  File "C:\Users\dominikschneider\Miniconda3\envs\test-environment\lib\multiprocessing\context.py", line 322, in _Popen
    return Popen(process_obj)
  File "C:\Users\dominikschneider\Miniconda3\envs\test-environment\lib\multiprocessing\popen_spawn_win32.py", line 46, in __init__
    prep_data = spawn.get_preparation_data(process_obj._name)
  File "C:\Users\dominikschneider\Miniconda3\envs\test-environment\lib\multiprocessing\spawn.py", line 172, in get_preparation_data
    main_mod_name = getattr(main_module.__spec__, "name", None)
AttributeError: module '__main__' has no attribute '__spec__'
```
Something didn't like the "

4. 

i replaced the space with \s and removed "   `(.{2})-(.+)-(\d{4}\d{2}\d{2}\s\d{2}\d{2}\d{2})-(.+)`
```
(test-environment) C:\Users\dominikschneider\Documents\phenomics\chickpea>run_workflows.cmd

python.exe C:\Users\dominikschneider\Miniconda3\envs\test-environment\Scripts\plantcv-workflow.py --dir data\slowgrowth --workflow vis_workflow.py --type png --json slowgrowth.json --outdir output-images --adaptor filename --delimiter (.{2})-(.+)-(\d{4}\d{2}\d{2}\s\d{2}\d{2}\d{2})-(.+) --meta plantbarcode,measurementlabel,timestamp,camera -T 6 -D 2019-10-05 -w --create --other_args="--debugdir debug-images"
Starting run 2019-10-17_15-49-28

Building job list...
Job list will include 4 images

took 0.02390289306640625

Processing images...
Traceback (most recent call last):
  File "C:\Users\dominikschneider\Miniconda3\envs\test-environment\Scripts\plantcv-workflow.py", line 4, in <module>
    __import__('pkg_resources').run_script('plantcv==3.6.1+6.g79605c9', 'plantcv-workflow.py')
  File "C:\Users\dominikschneider\Miniconda3\envs\test-environment\lib\site-packages\pkg_resources\__init__.py", line 666, in run_script
    self.require(requires)[0].run_script(script_name, ns)
  File "C:\Users\dominikschneider\Miniconda3\envs\test-environment\lib\site-packages\pkg_resources\__init__.py", line 1453, in run_script
    exec(code, namespace, namespace)
  File "c:\users\dominikschneider\miniconda3\envs\test-environment\lib\site-packages\plantcv-3.6.1+6.g79605c9-py3.7.egg\EGG-INFO\scripts\plantcv-workflow.py", line 319, in <module>
    main()
  File "c:\users\dominikschneider\miniconda3\envs\test-environment\lib\site-packages\plantcv-3.6.1+6.g79605c9-py3.7.egg\EGG-INFO\scripts\plantcv-workflow.py", line 292, in main
    pcvp.multiprocess(jobs, args.cpu)
  File "c:\users\dominikschneider\miniconda3\envs\test-environment\lib\site-packages\plantcv-3.6.1+6.g79605c9-py3.7.egg\plantcv\parallel\multiprocess.py", line 15, in multiprocess
    p = mp.Pool(processes=cpus)
  File "C:\Users\dominikschneider\Miniconda3\envs\test-environment\lib\multiprocessing\context.py", line 119, in Pool
    context=self.get_context())
  File "C:\Users\dominikschneider\Miniconda3\envs\test-environment\lib\multiprocessing\pool.py", line 176, in __init__
    self._repopulate_pool()
  File "C:\Users\dominikschneider\Miniconda3\envs\test-environment\lib\multiprocessing\pool.py", line 241, in _repopulate_pool
    w.start()
  File "C:\Users\dominikschneider\Miniconda3\envs\test-environment\lib\multiprocessing\process.py", line 112, in start
    self._popen = self._Popen(self)
  File "C:\Users\dominikschneider\Miniconda3\envs\test-environment\lib\multiprocessing\context.py", line 322, in _Popen
    return Popen(process_obj)
  File "C:\Users\dominikschneider\Miniconda3\envs\test-environment\lib\multiprocessing\popen_spawn_win32.py", line 46, in __init__
    prep_data = spawn.get_preparation_data(process_obj._name)
  File "C:\Users\dominikschneider\Miniconda3\envs\test-environment\lib\multiprocessing\spawn.py", line 172, in get_preparation_data
    main_mod_name = getattr(main_module.__spec__, "name", None)
AttributeError: module '__main__' has no attribute '__spec__'
```

not sure what else to try...

nfahlgren:
If I make a bat script like this it seems to work for me:

```
python.exe ^
%CONDA_PREFIX%\Scripts\plantcv-workflow.py ^
--dir data\slowgrowth ^
--workflow vis_workflow.py ^
--type png ^
--json slowgrowth.json ^
--outdir output-images ^
--adaptor filename ^
--delimiter "(.{2})-(.+)-(\d{4}\d{2}\d{2} \d{2}\d{2}\d{2})-(.+)" ^
--meta plantbarcode,measurementlabel,timestamp,camera ^
-T 6 ^
-D 2019-10-05 ^
-w ^
--create ^
--other_args="--debugdir debug-images"
```

dschneiderch:
unfortunately not working for me when i run the script. 
I stepped through `plantcv-workflow.py` though and it worked. Possibly a side effect of my options class - does it look ok? Replicating the args is making everything confusing.
```
class options():
    def __init__(self):
        self.dir = "data/slowgrowth"
        self.workflow = "vis_workflow.py"
        self.outdir = "output-images"
        self.json = "results.json"
        self.type = "png"
        self.adaptor = "filename"
        self.delimiter = "(.{2})-(.+)-(\d{4}\d{2}\d{2} \d{2}\d{2}\d{2})-(.+)"
        self.meta = "plantbarcode,measurementlabel,timestamp,camera"
        self.dates = False #does this default to False? the first use is if args.dates: so it must(?)
        self.match = None #but these default to None? very confused.
        self.coprocess = None
        self.create = False
        self.other_args = None
        self.writeimg = False
        self.cpu = 2
```

Looks like a problem with multiprocess? i tried cpu=1 and cpu=2 but never a problem interactively.

```
python.exe C:\Users\dominikschneider\Miniconda3\envs\test-environment\Scripts\plantcv-workflow.py --dir data\slowgrowth --workflow vis_workflow.py --type png --json slowgrowth.json --outdir output-images --adaptor filename --delimiter "(.{2})-(.+)-(\d{4}\d{2}\d{2} \d{2}\d{2}\d{2})-(.+)" --meta plantbarcode,measurementlabel,timestamp,camera
Starting run 2019-10-22_08-37-31

Building job list...
Job list will include 24 images

took 0.0329289436340332

Processing images...
Traceback (most recent call last):
  File "C:\Users\dominikschneider\Miniconda3\envs\test-environment\Scripts\plantcv-workflow.py", line 4, in <module>
    __import__('pkg_resources').run_script('plantcv==3.6.1+6.g79605c9.dirty', 'plantcv-workflow.py')
  File "C:\Users\dominikschneider\Miniconda3\envs\test-environment\lib\site-packages\pkg_resources\__init__.py", line 666, in run_script
    self.require(requires)[0].run_script(script_name, ns)
  File "C:\Users\dominikschneider\Miniconda3\envs\test-environment\lib\site-packages\pkg_resources\__init__.py", line 1462, in run_script
    exec(code, namespace, namespace)
  File "c:\users\dominikschneider\miniconda3\envs\test-environment\lib\site-packages\plantcv-3.6.1+6.g79605c9.dirty-py3.7.egg\EGG-INFO\scripts\plantcv-workflow.py", line 319, in <module>
    main()
  File "c:\users\dominikschneider\miniconda3\envs\test-environment\lib\site-packages\plantcv-3.6.1+6.g79605c9.dirty-py3.7.egg\EGG-INFO\scripts\plantcv-workflow.py", line 292, in main
    pcvp.multiprocess(jobs, args.cpu)
  File "c:\users\dominikschneider\miniconda3\envs\test-environment\lib\site-packages\plantcv-3.6.1+6.g79605c9.dirty-py3.7.egg\plantcv\parallel\multiprocess.py", line 15, in multiprocess
    p = mp.Pool(processes=cpus)
  File "C:\Users\dominikschneider\Miniconda3\envs\test-environment\lib\multiprocessing\context.py", line 119, in Pool
    context=self.get_context())
  File "C:\Users\dominikschneider\Miniconda3\envs\test-environment\lib\multiprocessing\pool.py", line 176, in __init__
    self._repopulate_pool()
  File "C:\Users\dominikschneider\Miniconda3\envs\test-environment\lib\multiprocessing\pool.py", line 241, in _repopulate_pool
    w.start()
  File "C:\Users\dominikschneider\Miniconda3\envs\test-environment\lib\multiprocessing\process.py", line 112, in start
    self._popen = self._Popen(self)
  File "C:\Users\dominikschneider\Miniconda3\envs\test-environment\lib\multiprocessing\context.py", line 322, in _Popen
    return Popen(process_obj)
  File "C:\Users\dominikschneider\Miniconda3\envs\test-environment\lib\multiprocessing\popen_spawn_win32.py", line 46, in __init__
    prep_data = spawn.get_preparation_data(process_obj._name)
  File "C:\Users\dominikschneider\Miniconda3\envs\test-environment\lib\multiprocessing\spawn.py", line 172, in get_preparation_data
    main_mod_name = getattr(main_module.__spec__, "name", None)
AttributeError: module '__main__' has no attribute '__spec__'
```

batch script:
```
python.exe ^
%CONDA_PREFIX%\Scripts\plantcv-workflow.py ^
--dir data\slowgrowth ^
--workflow vis_workflow.py ^
--type png ^
--json slowgrowth.json ^
--outdir output-images ^
--adaptor filename ^
--delimiter "(.{2})-(.+)-(\d{4}\d{2}\d{2} \d{2}\d{2}\d{2})-(.+)" ^
--meta plantbarcode,measurementlabel,timestamp,camera
```

I did however find a bug in `plantcv-workflow.py` that might be triggering something downstream with your changes?
you are creating a dictionary, regardless.
```
# Metadata restrictions
    args.imgtype = {}
    if args.match is not None:
        pairs = args.match.split(',')
        for pair in pairs:
            key, value = pair.split(':')
            args.imgtype[key] = value
    else:
        args.imgtype['None'] = 'None'
```
 when there is no `match` input 
https://github.com/danforthcenter/plantcv/blob/79605c9d330ec1bb1e30912de497d1fd067137eb/plantcv-workflow.py#L218 


but on L253 you have `if args.imgtype is not None` 
https://github.com/danforthcenter/plantcv/blob/79605c9d330ec1bb1e30912de497d1fd067137eb/plantcv-workflow.py#L253
 
which will always trigger True because args.imgtype is a dictionary and not `None`.
My `prefix` value is always `plantcv_NoneNone_....`

This would work:
`if next(iter(args.imgtype)) != 'None':`

I am still processing how this will interact downstream in `pcvp.metadata_parser()`. Does `args.imgtype` need to be a dictionary?  If not, then `args.imgtype=None` might be cleaner rather than `args.imgtype['None'] = 'None'`, and in any case we probably mean `args.imgtype['None'] = None`

I made 2 changes to my plantcv-workflow but it still didn't run. I'm still wrapping my head around `pcvp.metadata_parser` though. 
changed L218 to:
`args.imgtype['None'] = None`
and L253 to:
`if next(iter(args.imgtype)) !='None':`

maybe this helps?

dschneiderch:
@nfahlgren  another observation and thought:
it doesn't look like the date filter is working correctly. using this class below, the job_builder is giving me an entry for every image in the directory rather than just the last 4.
Could half the code be filtering for date and the other half not be filtered so things get misaligned?

```
class options():
    def __init__(self):
        self.dir = "data/slowgrowth"
        self.workflow = "vis_workflow.py"
        self.outdir = "output-images"
        self.json = "results.json"
        self.type = "png"
        self.adaptor = "filename"
        self.delimiter = "(.{2})-(.+)-(\d{4}\d{2}\d{2} \d{2}\d{2}\d{2})-(.+)"
        self.meta = "plantbarcode,measurementlabel,timestamp,camera"
        self.dates = "2010-10-05"
        self.match = "plantbarcode:D7"
        # self.match = None
        self.coprocess = None
        self.create = False
        self.other_args = None
        self.writeimg = False
        self.cpu = 2
        
args = options()
```

nfahlgren:
Hi @dschneiderch, not sure I'm following, is your options class just for trying to step through `plantcv-workflow.py` in a Jupyter notebook or something?

dschneiderch:
> 
> 
> Hi @dschneiderch, not sure I'm following, is your options class just for trying to step through `plantcv-workflow.py` in a Jupyter notebook or something?

correct

nfahlgren:
Ah ok, the options function in `plantcv-workflow.py` does some stuff with the inputs before returning the argparse object, but that is missing in your class.

nfahlgren:
It would probably be good if we modularized it as much as possible so it's covered in the testing environment but for now there's a fair amount of code in there.

nfahlgren:
I think this is why you are getting an error with multiprocessing: https://stackoverflow.com/questions/45720153/python-multiprocessing-error-attributeerror-module-main-has-no-attribute

dschneiderch:
> 
> 
> Ah ok, the options function in `plantcv-workflow.py` does some stuff with the inputs before returning the argparse object, but that is missing in your class.

I'm running that code after i define args = options()

nfahlgren:
I retested the date filter and it worked for me. The input has to be two datetimes in this format:

`--dates YYYY-MM-DD-hh-mm-ss_YYYY-MM-DD-hh-mm-ss`

dschneiderch:
> I think this is why you are getting an error with multiprocessing: https://stackoverflow.com/questions/45720153/python-multiprocessing-error-attributeerror-module-main-has-no-attribute

I get the error in the terminal too

dschneiderch:
> I retested the date filter and it worked for me. The input has to be two datetimes in this format:
> 
> `--dates YYYY-MM-DD-hh-mm-ss_YYYY-MM-DD-hh-mm-ss`

YYYY-MM-DD works too. i think any format that dateutil.parser() can handle.
turns out i was using 2010-10-05 instead of 2019-10-05 .... must be  quitting time.

dschneiderch:
so adding __spec__ = None just before calling main() seems to do the trick. 
https://stackoverflow.com/questions/45720153/python-multiprocessing-error-attributeerror-module-main-has-no-attribute
I still have no idea why it didn't work in the terminal after I reinstalled the dev version - maybe because something got corrupted from using jupyter kernel?

nfahlgren:
It seems to work on my Mac too, not sure if there's a consequence to setting `__spec__ = None`.

Looking back at the code you are correct, you can enter a single date and it will then keep images from that date until the current date (it sets the end date to now).

How about the regex though, that's still not working correct?

dschneiderch:
> It seems to work on my Mac too, not sure if there's a consequence to setting `__spec__ = None`.

sounds like it's usually `None` but maybe this means more to you https://docs.python.org/3/reference/import.html#special-considerations-for-main

> How about the regex though, that's still not working correct?

Now that I have that other issue resolved, i was able to mess with different regex patterns. In the format I used the above, both the regex and the simple '-' worked, so that's good. There is a problem when `_` are used, and more generally any character other than `-`  (I think).

It's complaining about the string `20190927 04_16_23`.   In `_check_date_range()` you have `dt_parser(img_time.replace("-", ""))` which doesn't convert the `_`.
Personally I think the best thing would be to add an argument for the dateformat in the regex, otherwise you will constantly be guessing what people are using. most likely the date separators will always be -, _,  or :  and you could squash all of them but who knows.  If the user provides `--dateformat %Y-%m-%d %H_%M_%S` then you could use datetime.datetime.strptime(timestamp, args.dateformat) and get it right every time.

```
import datetime
img_date = datetime.strptime('2019-10-05 04_16_17','%Y-%m-%d %H_%M_%S')
```
Additionally I think it would be worth considering saving img_meta['Timestamp'] as an ISO formatted date.


```
python.exe C:\Users\dominikschneider\Miniconda3\envs\test-environment\Scripts\plantcv-workflow.py --dir data\slowgrowth --workflow vis_workflow.py --type png --json slowgrowth.json --outdir output-images --adaptor filename --delimiter "(.{2})-(.+)-(\d{4}-\d{2}-\d{2} \d{2}_\d{2}_\d{2})-(.+)" --meta plantbarcode,measurementlabel,timestamp,camera --dates 2019-10-05 --match plantbarcode:D7
Starting run 2019-10-23_10-38-20

Traceback (most recent call last):
  File "C:\Users\dominikschneider\Miniconda3\envs\test-environment\Scripts\plantcv-workflow.py", line 4, in <module>
    __import__('pkg_resources').run_script('plantcv==3.6.1+10.g62800ab', 'plantcv-workflow.py')
  File "C:\Users\dominikschneider\Miniconda3\envs\test-environment\lib\site-packages\pkg_resources\__init__.py", line 666, in run_script
    self.require(requires)[0].run_script(script_name, ns)
  File "C:\Users\dominikschneider\Miniconda3\envs\test-environment\lib\site-packages\pkg_resources\__init__.py", line 1462, in run_script
    exec(code, namespace, namespace)
  File "c:\users\dominikschneider\miniconda3\envs\test-environment\lib\site-packages\plantcv-3.6.1+10.g62800ab-py3.7.egg\EGG-INFO\scripts\plantcv-workflow.py", line 321, in <module>
    main()
  File "c:\users\dominikschneider\miniconda3\envs\test-environment\lib\site-packages\plantcv-3.6.1+10.g62800ab-py3.7.egg\EGG-INFO\scripts\plantcv-workflow.py", line 274, in main
    coprocess=args.coprocess)
  File "c:\users\dominikschneider\miniconda3\envs\test-environment\lib\site-packages\plantcv-3.6.1+10.g62800ab-py3.7.egg\plantcv\parallel\parsers.py", line 200, in metadata_parser
    in_date_range = _check_date_range(start_date, end_date, img_meta['timestamp'])
  File "c:\users\dominikschneider\miniconda3\envs\test-environment\lib\site-packages\plantcv-3.6.1+10.g62800ab-py3.7.egg\plantcv\parallel\parsers.py", line 230, in _check_date_range
    timestamp = dt_parser(img_time.replace("-", ""))
  File "c:\users\dominikschneider\miniconda3\envs\test-environment\lib\site-packages\dateutil\parser\_parser.py", line 1358, in parse
    return DEFAULTPARSER.parse(timestr, **kwargs)
  File "c:\users\dominikschneider\miniconda3\envs\test-environment\lib\site-packages\dateutil\parser\_parser.py", line 649, in parse
    raise ValueError("Unknown string format:", timestr)
ValueError: ('Unknown string format:', '20190927 04_16_23')
```

dschneiderch:
better flag would be --timestampformat, -s so not to confuse with the (unspecified) format for  --dates, -D

nfahlgren:
You should be able to use `_` in a regex, it's not a protected character as far as I am aware.

nfahlgren:
Reformatting the image datetimes to [ISO 8601](https://en.wikipedia.org/wiki/ISO_8601) would be good, been on the list of things to do. The Python date objects have a `isoformat` method. Looks like it would output in the format `2019-09-27T04:16:23`. In my thinking we should also output in UTC, but I suspect dealing with figuring out local timezone might be difficult.

dschneiderch:
sure, we can for example replace `:`, `-`, `_`, ` `, `/`
```
timestamp = dt_parser(re.sub(pattern = '[\\:\\-_\\ \\/]',repl='', string = img_time))
```
but what happens when someone tries to use img_time = '05-10-2019'  for oct 5, 2019? this actually fails because we'd make it 05102019 and it tries to read it at YYYYMMDD.

also, 2019-10-05 04-16-17 for some reasons chops off part of the time.
```
In [65]: timestr = '2019/10/05 04-16-17'

In [67]: dt_parser(timestr, ignoretz=True)
Out[67]: datetime.datetime(2019, 10, 5, 4, 0)
```

just seems very prone to future errors.

nfahlgren:
I just meant that the regular expression option can accept `_`, you had mentioned you ran into a problem with it?

dschneiderch:
yep, I did and splitting the metadata from the filenames works great! But it doesn't know what to do with a timestamp with `_` in it. see regex below:

```
python.exe C:\Users\dominikschneider\Miniconda3\envs\test-environment\Scripts\plantcv-workflow.py --dir data\slowgrowth --workflow vis_workflow.py --type png --json slowgrowth.json --outdir output-images --adaptor filename --delimiter "(.{2})-(.+)-(\d{4}-\d{2}-\d{2} \d{2}_\d{2}_\d{2})-(.+)" --meta plantbarcode,measurementlabel,timestamp,camera --dates 2019-10-05 --match plantbarcode:D7
```

dschneiderch:
That should only affect the process if --dates is given, but i got the same error without --dates
Looks like the short circuiting in metadata_parser() isn't happening because args.end_date is being set to datetime.now().

nfahlgren:
Both start and end date have default values, so it actually does a date check no matter what. The start date is the beginning of Epoch time (0) and the end date is "now".

dschneiderch:
actually looks like 
        args.start_date = 1

I changed args.end_date to be None to fix metadata_parser so it doesn't check date range when no --dates are specified.

nfahlgren:
I think that works. The one thing it would sort of "break" is that it now would not work to supply a single date to the dates filter because end would be None. But that's probably sensible anyhow.

dschneiderch:
No, it works. end_date isn't called anywhere if it is None. I just tried to confirm.

nfahlgren:
It doesn't filter though right? I think we only check the date range when this is True:

```
if start_date and end_date and img_meta['timestamp'] is not None:
```

dschneiderch:
how do you want to handle the date format?
as far as I can tell from testing, there is no universal separator, or JUMP, for both date and time to be parsed correctly.
the regex is awesome but the downside is that it opens up the floodgates for strangely formatted dates.

dschneiderch:
yes it filters. I only set args.end_date to None if --date is not present.
So, in plantcv-workflow.py
```
if args.dates:
    ...   
   args.start_date = (start_td.days * 24 * 3600) + start_td.seconds
   args.end_date = (end_td.days * 24 * 3600) + end_td.seconds 
else:
        args.start_date = 1
        args.end_date = None

```

here is the proof:
```
(test-environment) C:\Users\dominikschneider\Documents\phenomics\chickpea>run_workflows.cmd

python.exe C:\Users\dominikschneider\Miniconda3\envs\test-environment\Scripts\plantcv-workflow.py --dir data\slowgrowth --workflow vis_workflow.py --type png --json slowgrowth.json --outdir output-images --adaptor filename --delimiter "(.{2})-(.+)-(\d{4}-\d{2}-\d{2} \d{2}_\d{2}_\d{2})-(.+)" --meta plantbarcode,measurementlabel,timestamp,camera --match plantbarcode:D7 --dates 2019-10-07
Starting run 2019-10-23_14-48-56

Building job list...
Job list will include 2 images

took 0.023936986923217773

Processing images...
took 20.02771496772766

Processing results...
took 0.37701869010925293
```

nfahlgren:
All looks good on my end. My inclination is to merge this PR and create a new one for the datetime format option for #473. I think converting to using a config file in #470 will take a bit longer, so for the short term we could just add a new flag.

dschneiderch:
well it still breaks with _ so I added a replacement for _ to "" since my lemnatec timestamps opened this can of worms. now at least the default lemnatec timestamps are possible. see #473 for the planned improvment

PR #457, opened 2019-10-10
Doc updates

dschneiderch:
**Describe your changes**
More doc fixes.
In the case of -T in the plantcv-workflow.py I changed the "threads" argument to actually be threads instead of cpus since that fits with -T better.


nfahlgren:
Hi @dschneiderch,

Question for you. Not sure if this matters to anyone but technically we don't use threads for parallelization, we use the Python multiprocessing interface instead (https://docs.python.org/3.7/library/multiprocessing.html). The change you propose is the easiest and I think it's recognizable to people. The alternative we could do is:

```
Workflow script file: -w, --workflow 
Parallel processes: -p, --processes
Save analysis images: -W, --writeimg
```

Not sure if it's worth the extra changes, I defer to what you prefer.

dschneiderch:
@nfahlgren since you bring it up, I think all those flags are quite confusing. In particular, its confusing that the same letter can mean different things in my workflow vs plantcv-workflow. Anything to standardize them and make them more intuitive would make it easier for people to use.

in VIS workflow:
```
-i, --image = Input image file
-o, --outdir = output directory
-r, --result = result file
-w, --writeimg = write out image (although you could use this as a T/F flag for anything)
-D, --debug = debug state
```

in parallel:
```
-d, --dir = directory of input images
-p, --workflow = workflow to use
-i, --outdir = output directory
-D, --dates = dates to select
```

so in particular, the `d` s, `i` s, and `o` s
I think it would be more intuitive if they were universal where possible:
```
-i,--input (this is an image in vis-workflow, a directory for parallel workflow)
-d, --dates = dates to select
-o, --outdir = output directory
-D, --debug = debug state (workflows only)
-O, --debugout = debug_outdir (workflows only)
```

I will also note that `sample_images` uses another set of i,o flags:
```
-s, --source = source or input directory
-o, --outdir = output directory
```
I would replace `-s` with `-i`

finally, yes, I think these would be good (although I was going to ask about the above how you felt about having them be case sensitive.) Personally I like to use the  flagnames so I don' t have to remember what each letter means. The benefit of `-p, --processes` is if you ever want to add `-T, --threads` then its available. processes is recognizable IMHO.
```
Workflow script file: -w, --workflow 
Parallel processes: -p, --processes
Save analysis images: -W, --writeimg
```

last, i find `-c`, `--create` a little awkward. at first I thought that it wouldn't make the results file unless i had `-c`. This is really  `--overwrite` if I understand correctly. `-f` for force since we already use `-o`? - but then `--meta` might have to become `-m`

HaleySchuhl:
I think now would be as appropriate a time for an overhaul as any! There have been a handful of changes to parallelization throughout v3 development so before we finish v3 we could reexamine flags, including ones that we plan to add in the near future, and try to make it as intuitive as possible. Documentation about old flags versus their new counterparts should go into the `updating.md` documentation page so existing users can update jobs without flipping between versions of documentation.

dschneiderch:
one more thing, I'd love to be able to specify a -l or -L  for --logdir.   the error files and run folders are getting overwhelming (and this is after I deleted error logs already)!
Screenshot_1

dschneiderch:
updated doc to change conda channel order. numerical packages changed and broke opencv somewhere. now prioritizing defaults channel which might be more conservative.  closes #469

dschneiderch:
I reverted the -T to --threads instead of --cpu and made a new issue for updating all flags #470 
Now all commits are just doc updates for clarity or to match the code in the function.

PR #459, opened 2019-10-10
Sample images fix

dschneiderch:
**Describe your changes**
Coerce 'num' argument in run_sample_images to integer to avoid `TypeError: '>' not supported between instances of 'str' and 'int'`

If i run sample_samples without an n argument it works fine. I could not figure out why or at what point num is getting converted to a str but that's what the error suggests. I tested the function with `num = int(num)` and it works for me.

```
(plantcv) C:\Users\dominikschneider\Documents\phenomics\GoldStandard2>python %CONDA_PREFIX%/Scripts/plantcv-utils.py sample_images -s data/raw_snapshots/vis -o data/testimages/vis -n 50
Traceback (most recent call last):
  File "C:\Users\dominikschneider\Miniconda3\envs\plantcv/Scripts/plantcv-utils.py", line 74, in <module>
    main()
  File "C:\Users\dominikschneider\Miniconda3\envs\plantcv/Scripts/plantcv-utils.py", line 69, in main
    options()
  File "C:\Users\dominikschneider\Miniconda3\envs\plantcv/Scripts/plantcv-utils.py", line 44, in options
    args.func(args)
  File "C:\Users\dominikschneider\Miniconda3\envs\plantcv/Scripts/plantcv-utils.py", line 58, in run_sample_images
    plantcv.utils.sample_images(source_path=args.source, dest_path=args.outdir, num=args.number)
  File "C:\Users\dominikschneider\Miniconda3\envs\plantcv\lib\site-packages\plantcv\utils\sample_images.py", line 61, in sample_images
    if num > len(img_element_array):
TypeError: '>' not supported between instances of 'str' and 'int'
```

**Type of update**
Is this a:
* Bug fix

**Additional context**
WIndows 10, bioconda plantcv v3.6

### DIFF
--- a/plantcv-utils.py
+++ b/plantcv-utils.py
@@ -29,7 +29,7 @@ def options():
     sample_images_cmd.add_argument("-s", "--source", help="Source directory of images", required=True)
     sample_images_cmd.add_argument("-o", "--outdir", help="Output directory for the random sample to get saved",
                                    required=True)
-    sample_images_cmd.add_argument("-n", "--number", help="The number of images to sample", default=100)
+    sample_images_cmd.add_argument("-n", "--number", help="The number of images to sample", default=100, type = int)
     sample_images_cmd.set_defaults(func=run_sample_images)
 
 
--- a/plantcv/utils/sample_images.py
+++ b/plantcv/utils/sample_images.py
@@ -8,10 +8,8 @@ def sample_images(source_path, dest_path, num=100):
     if not os.path.exists(source_path):
         raise IOError("Directory does not exist: {0}".format(source_path))
 
-    if not os.path.exists(dest_path):
-        os.mkdir(dest_path)
+    os.makedirs(dest_path, exist_ok=True)#python >=3.2
 
-    num = int(num)
     img_element_array = []
     sample_array = []
     num_images = []
@@ -45,8 +43,7 @@ def sample_images(source_path, dest_path, num=100):
             out_file.write(','.join(element))
             snap_path = os.path.join(source_path, "snapshot" + element[1])
             folder_path = os.path.join(dest_path, "snapshot" + element[1])
-            if not os.path.exists(folder_path):
-                os.mkdir(folder_path)
+            os.makedirs(folder_path, exist_ok=True)
             for root, dirs, files in os.walk(snap_path):
                 for file in files:
                     shutil.copy(os.path.join(root, file), folder_path)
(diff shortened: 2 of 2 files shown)


HaleySchuhl:
Hi @dschneiderch ,

Could you try a different patch for the bug that comes up with Windows while using the `sample_images` function? 

Does your error go away if you specify the type and change line 32 in `plantcv-utils.py` to `    sample_images_cmd.add_argument("-n", "--number", help="The number of images to sample", default=100, type=int)`?

dschneiderch:
> 
> 
> Hi @dschneiderch ,
> 
> Could you try a different patch for the bug that comes up with Windows while using the `sample_images` function?
> 
> Does your error go away if you specify the type and change line 32 in `plantcv-utils.py` to ` sample_images_cmd.add_argument("-n", "--number", help="The number of images to sample", default=100, type=int)`?

Yep, that works. see changes. 
Also, I made the swapped os.mkdir with os.makedirs because the latter is recursive. python 3.2 and up only though.

HaleySchuhl:
Since we still support Python 2.7 we might need to hold off on the second update until we drop Python 2 (In January? 🤞🏼)

dschneiderch:
I can not replicate the error from Travis on my computer. I tried various `dest_path` including with `..`, `.cache`.  Is it a unix thing or something about the CI setup?

nfahlgren:
Ah, figured out what is going on. In Linux/Unix the path starts with '/' so the first item in `dirs` is an empty string. Then in the `dirs` slice the range needs to be extended because the first slice, `dirs[0:i+1]` is `dirs[0:1]`, where 1 is not inclusive so it's equivalent to `dirs[0]` which is ''.

dschneiderch:
that won't work because dirs[0:i+2] on windows will contain 2 levels of directories and os.mkdir doesn't make new directories recursively.

dschneiderch:
turns out `os.makedirs()` is available in py2.7, just without the exist_ok argument to make it fail safe when the dir already exists.

nfahlgren:
I'm not sure how to add a test to cover the try/except block. It looks to me like we don't need it because we check if the path exists first. My understanding of how `os.makedirs` works is you only get an error if the target directory exists (the final directory in the path). Since we only run `os.makedirs` when `os.path.exists` is False, the directory will always exist and therefore we should not be able to get an OSError exception.

Does that make sense or am I wrong?

dschneiderch:
Yep, makes sense. I was assuming it would throw an error for intermediate directories too. I guess I'm not sure why you'd ever need the error catching then. I suppose there are other errors like permissions for why it would fail.  I will modify.

PR #480, opened 2019-10-29
Fix plantcv-workflow help function

nfahlgren:
**Describe your changes**
The help description text for the new `timestampformat` option has "%" characters that are used as placeholders by `argparse`. They need to be escaped using "%%" otherwise `argparse` throws an error when the help menu is printed. 

**Type of update**
Bug fix

**Additional context**
Solution: https://stackoverflow.com/questions/20480018/valueerror-unsupported-format-character

### DIFF
--- a/plantcv-workflow.py
+++ b/plantcv-workflow.py
@@ -146,8 +146,8 @@ def options():
                         default=None)
     parser.add_argument("-s", "--timestampformat", 
                         help='a date format code compatible with strptime C library, '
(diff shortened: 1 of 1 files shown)


dschneiderch:
any chance we can get 3.6.3 with this?

HaleySchuhl:
@dschneiderch 

The plan is to do a v3.7 release (hopefully this week?) with these updates and the beginning of the hyperspectral sub-package 😺

PR #504, opened 2020-01-02
Retain debug mode

### LINKED ISSUE #503: Fatal errors reset debugging mode
Many PlantCV functions use other functions from the pcv package. In order to avoid automatic intermediate debugging images getting output we store the debugging mode and reset it once we have created the debugging image for a particular function. When a function trips a fatal error it will also reset the debug mode to None since it never hits the part of the function that will reset it. This means that users have to rerun the line where they set the debugging mode after each time a function fails. 

Whenever there is a fatal error, reset the debug mode first. 

**Additional context**
This is only an inconvenience while building and testing a workflow but shouldn't have any impact on running a workflow over images in parallel.

HaleySchuhl:
**Describe your changes**
Read through all PlantCV functions and removed unused import statements, commented out legacy code, and optimized the storage and disabling of the debugging mode. 

**Type of update**
Is this a:

**Associated issues**
* closes #503


plantcv/plantcv/acute_vertex.py
```diff
@@ -24,12 +24,17 @@ def acute_vertex(img, obj, win, thresh, sep):
...
+    Returns:
+    acute_points = list of acute points
+    img2         = debugging image
+
     :param img: ndarray
     :param obj: ndarray
     :param win: int
     :param thresh: int
     :param sep: int
     :return acute: ndarray
```
nfahlgren:
```suggestion
    :return acute_points: ndarray
```

PR #505, opened 2020-01-09
Add threshold method for masking saturated pixels

### LINKED ISSUE #119: Mask saturated pixels
Saturated pixels (overexposed) can be a problem because they potentially underestimate the true pixel intensity. This can be a problem later when analyzing color data. Proposed solution:

Create a new function that detects saturated pixels (could be defined as all R = B = G = 255, or alternatively R = 255 OR B = 255 OR G = 255, etc.) and outputs a binary mask indicating which pixels are saturated or not. The saturated mask would be combined with the plant contour mask to remove the saturated pixels. The combined mask would be the mask input to `pcv.analyze_color` so that the output histograms would exclude saturated pixels.

bganglia:
**Describe your changes**
This pull request adds a function for masking saturated pixels. The function allows the user to choose one of two definitions of saturation:
  1. Any channel at or above a certain threshold
  2. All channels at or above a certain threshold.

The user can also pick this threshold (default = 255).

**Type of update**
Is this a:

**Associated issues**
Closes #119

**Additional context**
I am new to this project, so I am open to your suggestions on style/approach. Here are the decisions I made:
- To be consistent with the other code I have seen in this repo, this function mostly calls on openCV functions, instead of numpy
- I sometimes used openCV functions instead of plantcv functions, so that
  - There would be just one one debug image
  - The debug image would have saturation_mask in the name.

**Checklist**
- [x] Add documentation
- [x] Add tests


HaleySchuhl:
Thanks @bganglia ! The new function looks great. 

I can add unit tests and documentation to your branch if that's alright with you :) 

Also, in functions where we utilize other plantcv functions we store the debugging mode and temporarily disable it to avoid extra debugging images.

bganglia:
@HaleySchuhl 

Ok, that's all right with me. Also, I could shorten the function by using pcv.threshold.binary and pcv.threshold.custom_range, if you think that is a good idea.

Do you mind if I make a PR adding the note about changing and storing the debug setting to the "New function style guide" section of the contribution guide?

HaleySchuhl:
@bganglia I think the way you've written the function looks good! Using PlantCV functions internally can be really useful, especially since not all packages play nice together and we've usually done the legwork for converting to appropriate datatypes, but too much interdependencies has given us import order errors so I will weigh these considerations while testing an algorithm. 

You are more than welcome to open another PR for updating the contributing guide. Thanks for your contributions thus far!

tests/tests.py
```diff
@@ -4686,6 +4686,32 @@ def test_plantcv_threshold_custom_range_bad_input_channel():
...
+    _ = pcv.threshold.saturation(rgb_img=rgb_img, threshold=254, channel="all")
+    # Test with debug = "plot"
+    pcv.params.debug = "plot"
+    thresh = pcv.threshold.saturation(rgb_img=rgb_img, threshold=254, channel="any")
+    assert np.sum(thresh) == 920050455 and len(np.unique(thresh)) == 2
+
+
+def test_plantcv_threshold_saturation_bad_input():
+    # Test cache directory
+    cache_dir = os.path.join(TEST_TMPDIR, "test_plantcv_threshold_triangle")
```
nfahlgren:
```suggestion
    cache_dir = os.path.join(TEST_TMPDIR, "test_plantcv_threshold_saturation_bad_input")
```

tests/tests.py
```diff
@@ -4686,6 +4686,32 @@ def test_plantcv_threshold_custom_range_bad_input_channel():
         _, _ = pcv.threshold.custom_range(img, lower_thresh=[0], upper_thresh=[2], channel='CMYK')
 
 
+def test_plantcv_threshold_saturation():
+    # Test cache directory
+    cache_dir = os.path.join(TEST_TMPDIR, "test_plantcv_threshold_triangle")
```
nfahlgren:
```suggestion
    cache_dir = os.path.join(TEST_TMPDIR, "test_plantcv_threshold_saturation")
```

PR #520, opened 2020-01-31
Match multiple metadata values

### LINKED ISSUE #461: 'match' multiple values of same metadata field
It would be great if I could specify multiple values to match for the same metadata field key. This is another way to subset the images to test the workflow on more than a couple images.

```
python %CONDA_PREFIX%/Scripts/plantcv-workflow.py --workflow vis.py ^
    --dir data/testimages/vis ^
    --type png ^
    --outdir output ^
    --json results.json ^
    --adaptor filename ^
    --delimiter - ^
    --meta plantbarcode-measurementlabel-timestamp-camera ^
    -T 4 ^
    --create ^
    --other_args="--debugdir debug-images" ^
    --match plantbarcode:A1,plantbarcode:B2,plantbarcode:C3
```

It would be particularly handy to filter the dates this way so I could run the first day and last day.

- an alternative image directory that has a subset of files
- run workflow multiple times with different match arguments

**Additional context**
plantcv-utils.py sample_images  could potentially be used as an intermediary if you could sample based on group. same idea as previously discussed with groups and maps.

bganglia:
**Describe your changes**
This PR allows selecting more than one value within each metadata field. For example, this would make it possible to process images from only the first and last dates of an experiment, or some subset of the cameras used.

**Type of update**
Is this a:
* Work in progress

**Associated issues**
This will close #461

**Additional context**
1. Data structures on the Python side
In this PR, meta_filters is a dictionary with keys that can be single values or lists. If they are lists, metadata_parser checks whether at least one matches. So this is a valid group of filters:
```
meta_filters = {
    "id":["11", "32"],
    "plantbarcode":"A1"
}
```
2. Syntax on the CLI side
Two different syntax styles were mentioned in the issue:
a. ``--match plantbarcode:A1,plantbarcode:B2,plantbarcode:C3`` This way is easier for writing CLI scripts that wrap plantcv-workflow.py and simpler to parse in Python.
b. ``--match plantbarcode:[A1,B2,C3]`` This way is much easier to read and probably better for experimentation, but harder to parse in Python.
In this PR, metadata_parser accepts either/both. For example, the following would all be read in the same way:
``--match id:[1,2,3]``
``--match id:1,id:2,id:3``
``--match id:[1,2],id:3``
If this is too much flexibility, the code could be limited to one style.

**To Do**
- [X] Accept multiple metadata values
- [X] Support ``\`` and quoting
- [x] Raise informative (fatal) error if invalid metadata filters are given
- [x] Refactor
- [x] Make sure that all current tests pass
- [x] Add new tests
- [ ] Add documentation


dschneiderch:
since you've done both I dont know why we wouldn't implement both. this is going to be great! Thanks @bganglia

bganglia:
@nfahlgren 
I have a few questions about error handling before I wrap up the tests for this feature. Right now the code assumes the user meant something reasonable instead of throwing warnings or errors when the metadata filters look fishy. Specifically:
1. Empty metadata values (``""``) are valid.
2. Presently, ``:`` and ``,`` are not interpreted as special characters when that would create an error. For example, ``id::`` would translate to ``{"id":":"}``. Alternately, you could require backslashes, meaning the user would have to write ``id:\:`` to avoid an error.

Which style is more fitting for PlantCV - throwing more errors/warnings or making more assumptions?

HaleySchuhl:
@bganglia , thanks for all your work on this so far! 

I think that the first example might make most sense to be handled as if no matching is needed for that type of metadata but also print a warning. For example if someone input `--match treatment:[]` I would think most people meant all treatments should be included rather than no images. 

The second case with handling special characters such as `:` I like your idea about allowing slashes since it is similar to the regex style we've been moving towards regarding timestamps. A fatal error with an informative error message would be best in my opinion in the case where the escape character is missing before special characters. Do you agree @nfahlgren

nfahlgren:
Yeah, I think if there are any errors we think would be common/likely it's worth putting some error handling around them, but I think it's okay to make some assumptions for everything else. Just comes down to the amount of effort in writing checks for lots of probably unlikely inputs, of which there could be just about anything. I think this is where the documentation helps, plus users posting issues when they have data that breaks our assumptions that we can then improve the code around.

It is possible people will need to use characters like `:` since it's a valid character in filenames in Linux/Unix, and that's where most metadata is extracted from. Having the option to escape sounds good to me, but maybe an easier alternative is to instruct users to quote the strings?

bganglia:
@HaleySchuhl @nfahlgren That makes sense. I will modify the checklist to include these ideas.

Just to be clear, should the syntax ``id:``, which is currently equivalent to ``id:[]``, raise an error or also be interpreted as selecting all images?

nfahlgren:
In my thinking it's an error. The default behavior is to not filter on any metadata, so all `id` in this case are included automatically. So there's no need for a user to specify that all `id` should be matched.

bganglia:
@nfahlgren OK. I will raise an error on empty filters instead.

dschneiderch:
this might be way too late and or irrelevant but I recently learned that argparse has a flag `nargs` to allow multiple inputs. so instead of 
`--match plantbarcode:A1,plantbarcode:B2`
you'd use
`--match plantbarcode:A1 plantbarcode:B2` and it would be read in as an array of inputs
I don't know if that would help simplify anything
https://docs.python.org/3/library/argparse.html

bganglia:
@dschneiderch Thank you. I have been looking for a way to simplify the parser code, and this could help a lot.

In that case, argparse would be responsible for separating each filter and PlantCV would only have to apply backslashing, quoting, and the bracketed list syntax (if applicable). 

There is only one side effect I can think of. Currently, all syntax errors lead to a message showing where the problem occurred, like this:
```
ValueError: Empty value
1:9,2:'',4:\:
       ^
```
I am not sure if would be possible to show the entire line using ``nargs``, but it might be. I will have to check into this

dschneiderch:
@bganglia Is this error happening as part of the `argparse`  or when you access `args.match`? for the latter you should be able to catch any ValueError and then show whatever you want to show.
https://wiki.python.org/moin/HandlingExceptions

bganglia:
@dschneiderch argparse accepts the input without throwing any errors--the error is raised on purpose by PlantCV because it does not know how to interpret the filter.

I meant that if the parser code I wrote starts with individual items (e.g. ``1:9`` ``2:''`` ``4:\:``) instead of the entire line, then the error message might be harder to understand:

```
ValueError: Empty value
2:''
  ^
```

But now that I have been looking into argparse a little bit more, it looks like parsing backslashes and quotes is a built-in feature, so that can actually be taken out of this code as well. I think that would leave just parsing the key-value pairs and bracketed list syntax.

bganglia:
After looking into things a little more, it seems like the longer error message will still work. 

Still, I cannot find any way to use ``,`` as the delimiter in argparse, so the filters would have to be separated by spaces and not commas. This could potentially break backwards compatibility.

bganglia:
Actually, after thinking about \ and "" a little more, I realized that, even though argparse already supports these, removing them from this parser would break cases with e.g. colons or commas in identifiers.

So I am leaving support for quoting and bracketing in,

bganglia:
@nfahlgren @HaleySchuhl 
This pull request is nearing completion, so I wanted to check in to make sure I am on the right track.

This PR adds the following:
1. Matching multiple metadata values specified either as ``id:1,id:2`` or ``id:[1,2]``
2. Support for quoting and escaping ``,``, ``[``, ``]``, and ``:``
3. Very specific and helpful error messages when syntax errors are present, e.g.
```
EmptyListError: Empty list
id:[]
    ^
```

Internally, the code splits the input into basic units like ``camera`` or ``:`` and has hard-coded rules to decide how to build the dictionary and recognize errors.

If I am on the right track, I will wrap up the documentation and mark this PR as completed. If not, I can try to change things or work on something else instead.

bganglia:
I think that there is a better approach than the one that I was using here, so I will open a new pull request to replace this one.

PR #529, opened 2020-02-24
roi2contour

HaleySchuhl:
**Describe your changes**
Add a function that allows user to create a binary mask from any contour. Also, add the option for whether or not to include the original image in the new `plantcv.visualize.colorspaces` function. The function also scales down the debugging image now too. 

**Type of update**
Is this a:
* New feature

**Associated issues**

**Additional context**


plantcv/plantcv/roi2mask.py
```diff
@@ -0,0 +1,38 @@
...
+    :param roi_contour: numpy.ndarray
+    :param contour: list
+    :return mask: numpy.ndarray
+    """
+    params.device += 1
+
+    # create a blank image of same size
+    shape_info = np.shape(img)
+    bnk = np.zeros((shape_info[0], shape_info[1]), dtype=np.uint8)
+    img1 = np.copy(img)
```
nfahlgren:
```suggestion
```

dschneiderch:
Oh, nice! I was just looking for a function like this. Nice for avoiding `roi_object`. The roi subpackage is where I looked for it but I only found `from_binary_image`

PR #526, opened 2020-02-19
Update hyper analyze index hist

HaleySchuhl:
**Describe your changes**
* Previously, the `plantcv.plantcv.hyperspectral.analyze_index()` function would mask an index/hyperspectral band from a specific wavelength and then calculate statistics about that masked region (e.g. mean, median, std). However, it was brought up that these statistics are smoothing over a lot more detail that can be extracted from these types of single-channel-images. Similar to the `plantcv.plantcv.analyze_nir()` function, the update now includes an optional histogram option that will save out frequencies of values within a masked region. We needed to include a `bin_max` and `bin_min` parameter since the natural range of pixel values can vary from image to image. Also updated documentation and testing to reflect updates. 
* Add min/max value method to `Spectral_data` class. This will eventually come in handy if we need to let power users adjust the range being used by `cv2.calcHist`. Made updates to functions to implement this new method but it's not fully being utilized just yet. 

**Type of update**
* Enhancement and new functionality added to the `pcv.hyperspectral.analyze_index()` function. 

**Associated issues**
#203  
closes #525

Issue #203, opened 2017-12-05
Hyperspectral submodule


nfahlgren:
### Description

@maliagehan has been working on some hyperspectral data-related functions for PlantCV. We discussed the possibility of making these tools optional since there may end up being dependencies that users not working with hyperspectral data may not want/need to install. Based on a discussion in the TERRA-REF project, we are proposing to develop the hyperspectral submodule (and possibly other submodules) as a separate GitHub repository (https://github.com/danforthcenter/plantcv-hyperspectral). We would design it and the main repository in such a way that you could install PlantCV by itself as is done now and optionally install the hyperspectral module into the PlantCV package namespace so that it would function like it's part of the package if it's installed. This way it's optional but we can still do continuous integration testing and everything in the linked repository. We could use this framework for other submodules that could/should be optional.

### Details

We will need to do some testing to figure out the best way to structure the submodule. We may need to refactor the main PlantCV repository somewhat to achieve the right structure.

### Completion Criteria

This is a discussion/proposal, specific issues should be opened for implementing the various components of this plan.

nfahlgren:
I want to briefly summarize a team meeting with @maliagehan, @HaleySchuhl and @josectovar.

Based on some testing we did, it looks like the simplest approach will be to read hyperspectral images directly with NumPy. This has the advantage of keeping the images in a format that is more easily compatible with existing PlantCV methods and does not increase dependency burden.

Given that, I propose that we actually put the hyperspectral submodule directly in the PlantCV main repository instead of the planned add-on package plan. It's going to be easier to work with the two packages if they are together and we can avoid duplicating CI and documentation infrastructure. We can always split them later if we decide we need to.

Thoughts?

HaleySchuhl:
Summary of today's team meeting:

* We each agree that it makes sense to add a hyperspectral sub-package to PlantCV repo rather than updating separate repositories and requiring users to download two packages. We also have been able to avoid additional dependencies right now. 
*  For now, don't worry about data types aside from ENVI. Until someone requests support for another type of image don't worry about trying to handle it. 
* Ideally, all functions in the hyperspectral sub-package should be able to handle different numbers of bands and different wavelength values.
* Eventually we would like to use classes to simplify programming for end users. For now, dump array that is the hypercube and add classes in as we go. 
[...]

PR #535, opened 2020-03-02
Analyze index march2020 updates

HaleySchuhl:
**Describe your changes**
Minimum and maximum bin parameters now control range of histogram values. There was a plotting mismatch between histogram plotting and calculation since the range for frequency values used to be independent of `min_bin` and `max_bin`. 

**Type of update**
Is this a:
* Bug fix
* Feature enhancement

**Associated issues**
#203 
#526

### DIFF
--- a/docs/analyze_index.md
+++ b/docs/analyze_index.md
@@ -3,7 +3,7 @@
 This function calculates the hyperspectral index statistics and writes the values  as observations out to
        the [Outputs class](outputs.md).
        
-**plantcv.hyperspectral.analyze_index**(*index_array, mask, bins=100, min_bin=None, max_bin=None*)
+**plantcv.hyperspectral.analyze_index**(*index_array, mask, bins=100, min_bin=0, max_bin=1*)
 
 **returns** None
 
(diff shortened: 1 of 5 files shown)


nfahlgren:
Wouldn't there still be an issue if the observed min and max for an index image is not the full index range?

HaleySchuhl:
@nfahlgren If the observed min and max are within 0 and the auto-calculated max then there's no problem. In the case where someone sets min=0 and max=1 based on theoretical values there's also no issue even if the observed range is not as wide as the theoretical range. In the case where the observed range is outside of the user defined `min_bin` and `max_bin` then the frequency data will be incomplete but still accurate.

nfahlgren:
Would each of the histograms be on a different scale though? For example, if a given set of NDVI images have a variety of observed max values (0.8, 0.9, 1, etc.). Though I guess the user could set the max.

HaleySchuhl:
@nfahlgren that's correct. Under the default settings there will likely be issues if people want to plot more than one on the same graph. I could update this a little more and have the defaults be `min_bin=0` (which is already the case but this is maybe more descriptive/accurate) and `max_bin=1` since *most* of the indices are expected to range from 0-1 assuming a biologically meaningful context. I could also print a warning statement if the case where observed min or max is outside the user defined range.

PR #544, opened 2020-03-11
fixed symbol causing error

jberry47:
Non-asci character was used and threw error:

SyntaxError: Non-ASCII character '\xe2' in file plantcv/plantcv/plantcv/hyperspectral/extract_index.py on line 85, but no encoding declared; see http://python.org/dev/peps/pep-0263/ for details

Replaced rogue character with ascii character


nfahlgren:
Note, this bug was causing an issue in Python 2.7, it doesn't like UTF-8 characters in the code (including comments).

PR #552, opened 2020-03-20
update param name from rgb_img to img

### LINKED ISSUE #546: Custom range threshold input variable requests RGB image but works with grayscale
**Describe the bug**
The main input argument for `pcv.threshold.custom_range` is `rgb_img`, but the function works with grayscale images also. This should be updated for clarity to `img`.

- [ ] update input variable name
- [ ] update tests as needed
- [ ] update documentation
- [ ] update the "Updating PlantCV" page in the docs

HaleySchuhl:
**Describe your changes**
The function `pcv.threshold.custom_range` had an input parameter named `rgb_img` which is inaccurate since the function also works on grayscale images. These updates correct the parameter name to the more general `img` in 

**Type of update**
Is this a:
* Bug fix
* Update to documentation

**Associated issues**
* closes #546

### DIFF
--- a/docs/custom_range_threshold.md
+++ b/docs/custom_range_threshold.md
@@ -1,14 +1,14 @@
 ## Custom Range Threshold
 
-Creates a binary image (mask) and a masked image from an RGB image based on the user defined threshold values. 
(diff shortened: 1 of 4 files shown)


DannieSheng:
@HaleySchuhl Your are right. I agree!

PR #558, opened 2020-04-03
Update features for extract index

DannieSheng:
**Describe your changes**
1. Move pcv.hyperspectral.extract_index to pcv.spectral_index
2. Adding features to the extract_index function in hyperspectral subclass:
a. Make the theoretical range for every VI inside the description of every function
3. Adding new index (EGI)

2. Fix wrong formulas and typos

**Type of update**
Is this a:
* Bug fix
* Update to documentation
* Work in progress

**Associated issues**

**Additional context**


DannieSheng:
> I pushed just a couple more lines of code where `pcv.hyperspectral.analyze_index` gets used during testing which was causing a couple of tests to fail. As long as I get those tests to pass I think it looks good to me!

Sounds good. Thank you Haley!

HaleySchuhl:
It looks like coveralls if failing since there are a few `if` statements that are never true with our current tests/testing data. One suggestion I have that will avoid writing new tests or including additional testing data would be to always cast into an array regardless of if the values are scalars or not. 

image

nfahlgren:
Maybe we could meet today to discuss our approach? I may not be thinking about all the various dimensions of the issue here, but my preference would be to modify the calculations to avoid the `nan` and `inf` issues rather than create a new data structure for keeping track of these pixels.

DannieSheng:
A meeting sounds good to me.

HaleySchuhl:
Sounds good to me too!

maliagehan:
sounds good o me as well.
_ _ _

Malia Gehan, Ph.D.
Assistant Member and Principal Investigator
Donald Danforth Plant Science Center
St. Louis, Mo, 63132
http:www.gehan-lab.org/
Office: 314-587-1420
Cell: 808-226-2399

nfahlgren:
Hi @DannieSheng, right now there is a `msg` variable with the range of each index but they are not used. We talked about printing the `msg`'s (and having a quiet mode in `pcv.params` to disable them). An alternative way we could make this information easily available to a user would be to split `plantcv.hyperspectral.extract_index` into a function per index. The index range could then just be in the function docstring so it's easy to look up with the help command or the shift+tab feature in Jupyter. It could be structured as a subpackage of `hyperspectral` called `index`, so each function would be called like this for example: `plantcv.hyperspectral.index.ndvi`, etc. We also wouldn't need all the if...elif code for determining whether the use input one VI name or another. What do you and @HaleySchuhl think?

DannieSheng:
Hi @nfahlgren, I am sorry in the earlier version the msg was printed out, however I accidentally deleted that in the latest pull request. I can add that back. Also, the idea of splitting  plantcv.hyperspectral.extract_index into a function per index sounds good to me. What do you think about?

nfahlgren:
No worries @DannieSheng! I think splitting them into separate functions would make it easier for a user to work with each index without having to reference the documentation as much. We can still keep the functions in one file so I don't think it adds too much extra complexity on the developer end either. Given those reasons I think it could be a good way to go. 

I'm thinking the setup would look something like this:

```
plantcv/
    hyperspectral/
        index/
            __init__.py
            spectral_indices.py
```

Then a small rearrangment of the code in the current extract_index.py module:

```python
def ndvi(hsi, distance=20):
    """Normalized Difference Vegetation Index.
    The theoretical range for NDVI is [-1,1]

    inputs:
    hsi           = hyperspectral image (PlantCV Spectral_data instance)
    distance = how lenient to be if the required wavelengths are not available

    Returns:
    index_array    = Index data as a Spectral_data instance
    :param hsi: __main__.Spectral_data
    :param distance: int
    :return index_array: __main__.Spectral_data
    """
    params.device += 1

    if (float(hsi.max_wavelength) + distance) >= 800 and (float(hsi.min_wavelength) - distance) <= 670:
        # Obtain index that best represents NIR and red bands
        nir_index = _find_closest(np.array([float(i) for i in wavelength_dict.keys()]), 800)
        red_index = _find_closest(np.array([float(i) for i in wavelength_dict.keys()]), 670)
        nir = (array_data[:, :, [nir_index]])
        red = (array_data[:, :, [red_index]])
        # Naturally ranges from -1 to 1
        index_array_raw = (nir - red) / (nir + red)
        return _package_index(hsi=hsi, raw_index=index_array_raw, method="NDVI")
    else:
        fatal_error("Available wavelengths are not suitable for calculating NDVI. Try increasing distance.")


def _package_index(hsi, raw_index, method):
    """Private function to package raw index array as a Spectral_data object.
    Inputs:
    hsi             = hyperspectral data (Spectral_data object)
    raw_index = raw index array
    method    = index method (e.g. NDVI)

    Returns:
    index        = index image as a Spectral_data object.
   
    :params hsi: __main__.Spectral_data
    :params raw_index: np.array
    :params method: str
    :params index: __main__.Spectral_data
    """
    # Reshape array into hyperspectral datacube shape
    index_array_raw = np.transpose(np.transpose(raw_index)[0])

    # Store debug mode
    debug = params.debug
    params.debug = None

    # Resulting array is float 32 from varying natural ranges, transform into uint8 for plotting
    all_positive = np.add(raw_index, 2 * np.ones(np.shape(raw_index)))
    scaled = rescale(all_positive)

    # Find array min and max values
    obs_max_pixel = float(np.nanmax(raw_index))
    obs_min_pixel = float(np.nanmin(raw_index))

    index = Spectral_data(array_data=raw_index, max_wavelength=0,
                                          min_wavelength=0, max_value=obs_max_pixel,
                                          min_value=obs_min_pixel, d_type=np.uint8,
                                          wavelength_dict={}, samples=hsi.samples,
                                          lines=hsi.lines, interleave=hsi.interleave,
                                          wavelength_units=hsi.wavelength_units, 
                                          array_type="index_" + method.lower(),
                                          pseudo_rgb=scaled, filename=hsi.filename, default_bands=None)

    # Restore debug mode
    params.debug = debug

    if params.debug == "plot":
        plot_image(index.pseudo_rgb)
    elif params.debug == "print":
        print_image(index.pseudo_rgb,
                    os.path.join(params.debug_outdir, str(params.device) + method + "_index.png"))

    return index
```

DannieSheng:
Sure, how about create a new pull request for this change?

nfahlgren:
Could keep the changes on this branch/pull request since it would replace the existing code being updated here. Almost all the code already in this PR would still be used, including the tests, just rearranged a bit.

DannieSheng:
Sure, that makes sense.

HaleySchuhl:
Finally just got all caught up with this thread! Sounds good to me :) @DannieSheng just slack me and remove the "work in progress" tag when this PR is all set for review 😁I can go through and test it pretty much whenever today or tomorrow.

DannieSheng:
@HaleySchuhl Sure, thank you Haley!

plantcv/plantcv/spectral_index/spectral_index.py
```diff
@@ -0,0 +1,863 @@
...
+        red_index = _find_closest(np.array([float(i) for i in hsi.wavelength_dict.keys()]), 670)
+        nir = (hsi.array_data[:, :, [nir_index]])
+        red = (hsi.array_data[:, :, [red_index]])
+        # Naturally ranges from -1 to 1
+        index_array_raw = (nir - red) / (nir + red)
+        return _package_index(hsi=hsi, raw_index=index_array_raw, method="NDVI")
+    else:
+        fatal_error("Available wavelengths are not suitable for calculating NDVI. Try increasing distance.")
+
+def gdvi(hsi, distance=20):
```
nfahlgren, line 41: GDVI is nir - green, not red. The Sripada et al. paper used G = 520-620, and TERRA-REF uses R580. I think we meant to use R580 here? Since the equation is just nir - green, I'm not sure I see why the range is -2.0 to 2.0.

plantcv/plantcv/spectral_index/spectral_index.py
```diff
@@ -0,0 +1,863 @@
...
+    if (float(hsi.max_wavelength) + distance) >= 570 and (float(hsi.min_wavelength) - distance) <= 531:
+        # Obtain index that best approximates 570 and 531 nm bands
+        pri570_index = _find_closest(np.array([float(i) for i in hsi.wavelength_dict.keys()]), 570)
+        pri531_index = _find_closest(np.array([float(i) for i in hsi.wavelength_dict.keys()]), 531)
+        pri570 = (hsi.array_data[:, :, [pri570_index]])
+        pri531 = (hsi.array_data[:, :, [pri531_index]])
+        # PRI = (R531- R570)/(R531+ R570))
+        denominator = pri531 + pri570
+        # Avoid dividing by zero
+        index_array_raw = np.where(denominator == 0, 0, ((pri531 - pri570) / denominator))
```
nfahlgren, line 114: Looks like in PRI we are avoiding division by zero but we are not doing that in the indices above. Since we are going to handle `nan` outputs later, I think we should just make this a simple equation like the other indices.

plantcv/plantcv/spectral_index/spectral_index.py
DannieSheng: Sure, since these are already there in plantcv so I did not check the references. I have made the changes.

plantcv/plantcv/spectral_index/spectral_index.py
DannieSheng: Sure, I have made the changes.

docs/spectral_index.md
```diff
@@ -0,0 +1,786 @@
...
+savi_array  = pcv.spectral_index.savi(hsi=spectral_data, distance=20)
+
+# Extract ARI index from the datacube
+ari_array  = pcv.spectral_index.ari(hsi=spectral_data, distance=20)
+
+# Extract CI_REDEDGE index from the datacube 
+ci_rededge_array  = pcv.spectral_index.ci_rededge(hsi=spectral_data, distance=20)
+
+# Extract CRI1 index from the datacube 
+cri1_array  = pcv.hyperspectral.extract_index.cri550(hsi=spectral_data, distance=20)
```
HaleySchuhl, line 606: remove

PR #559, opened 2020-04-07
custom_range threshold in RGB

### LINKED ISSUE #556: custom_threshold rgb implementation/documentation
**Describe the bug**
The implementation / documentation for custom_threshold is very confusing.

The channel is 'RGB' but actually you have to provide 'BGR'.  I'm inclined to change the implementation to RGB for consistency with the rest of the user facing functions but that would break existing code.  Alternatively we could change the documentation to 'BGR' and highlight the order of the channels explicitly.

@HaleySchuhl @nfahlgren  what do you think?

dschneiderch:
**Describe your changes**
Changed the method to interpret input in RGB mode as R, G, B instead of B,G,R as discussed in #556 

**Type of update**
Is this a:
* Bug fix

**Associated issues**
closes #556 

**Additional context**
**this is a breaking change but will be better in the long run in terms of consistency.** the documentation did not need changing because it implied R,G,B and the examples used the same threshold value for R and B channels.
Tests did not need updating because the thresholds are the same across the channels.


dschneiderch:
sorry, these pull requests got all screwed up. i should have branched both off master instead of off one another.
#560 I thought the example script at the bottom was redundant to the one above it. I couldn't figure out the difference so I removed it siince it confused me.

nfahlgren:
I think until we can check we want to assume the input image is an OpenCV image in BGR order. But the RGB upper/lower thresholds should be in RGB order and we apply the right order on the backend. That would mean we would leave the image subsetting section the same and change the inRange lines instead:

```python
# Separate channels (pcv.readimage reads RGB images in as BGR)
blue = img[:, :, 0]
green = img[:, :, 1]
red = img[:, :, 2]

# Make a mask for each channel
r_mask = cv2.inRange(red, lower_thresh[0], upper_thresh[0])
g_mask = cv2.inRange(green, lower_thresh[1], upper_thresh[1])
b_mask = cv2.inRange(blue, lower_thresh[2], upper_thresh[2])
```

dschneiderch:
@nfahlgren  good catch. same result but my way was unnecessarily confusing.

PR #568, opened 2020-04-15
Add function to threshold method

DannieSheng:
**Describe your changes**
A function called "mask_bad" is added to plantcv/plantcv/threshold/threshold_methods.py. This function is useful for creating a mask based on input grayscale image and a dictionary of indices of user defined "bad" pixels.

**Type of update**
Is this a:
* Update to documentation
* Work in progress

**Additional context**
This pull request follows another one, #558, and will followed by another one, that used the output of this function.


nfahlgren:
Hi @DannieSheng, rather than make a dictionary of bad pixels, what if we just use NumPy functions to make the mask in the `mask_bad` function?

```python
mask = np.zeros(np.shape(gray_img), dtype=np.uint8)
if bad_type.lower() == "native":
    mask[np.isnan(gray_img)] = 255
    mask[np.isinf(gray_img)] = 255
elif bad_type.lower() == "inf":
    mask[np.isinf(gray_img)] = 255
elif bad_type.lower() == "nan":
    mask[np.isnan(gray_img)] = 255
else:
    fatal_error("Not a valid bad_type: {}".format(bad_type))
```

DannieSheng:
@nfahlgren That makes much sense! Thank you, I will change that!

DannieSheng:
@nfahlgren Actually that would depend on what we want the function to do. The my function is one that takes a dictionary of use defined "bad" pixels that we want to mark. The dictionary is an output of the extract_index function, which indicates those pixels with some irregular values in some bands of their spectra. Which means that contains the information about the hyper spectral image and we can show these pixels in the pseudo-color visualization. My understanding is that it is a little bit different from taking an gray image and mark the "bad" based on that gray image.

nfahlgren:
@DannieSheng in the modified `extract_index` function the dictionary is created using the NumPy `isnan` and `isinf` functions to find the locations of these values in the grayscale VI image. But the VI image containing nan and inf values already has the location information encoded in the VI array. So rather than building a dictionary, we can use the NumPy `isnan` and `isinf` functions to make the mask in your new bad_mask function. It looks to me like it achieves the same functionality you are looking for but without a new output to `extract_index` and with one fewer input to `mask_bad` correct?

DannieSheng:
@nfahlgren Thank you. I see and I agree. So I think for the extract_index function, we need only one output, which is the VI. And I will have to change this mask_bad function. Is everything correct?

nfahlgren:
Sorry for the delay @DannieSheng, that sounds great!

nfahlgren:
Hi @DannieSheng,

Right now in `mask_bad` it will trigger a fatal error if a user asks to mask `nan` or `inf` but their image does not have those values. Couldn't we just return a black mask? The fatal error will trigger your workflow to stop, so you lose any additional analysis you might want to do.

DannieSheng:
@nfahlgren Your are right! I agree. I have updated the function and right now travisCI is running the test.

DannieSheng:
This can be closed after we merge #569

PR #576, opened 2020-05-19
Update read envi

### LINKED ISSUE #565: Set default wavelength units for ENVI files
@Mamatemenrs gave us some sample hyperspectral data to test with because it was not working with `pcv.readimage`. I found two issues.

First, the file names were:

```
REFLECTANCE_2019-06-09_010.dat
REFLECTANCE_2019-06-09_010.hdr
```

But we expect:

```
REFLECTANCE_2019-06-09_010.dat
REFLECTANCE_2019-06-09_010.dat.hdr
```

Second, the `hdr` file was missing the `wavelength units` key-value pair.

For the missing `wavelength units` key in the `hdr` file, I think we can check if the key exists in the parsed header file and if it doesn't set `wavelength units` to `nm` by default. If I rename the files as we expect and add this key-value pair to the `hdr` file PlantCV reads the image in correctly.

For the file extensions, I think what we have and what @Mamatemenrs has seem like they could both be common. Maybe we should check for the existence of `filename + ".hdr"`, and if it doesn't exist, check for `filename[:-4] + ".hdr"`?

HaleySchuhl:
**Describe your changes**
* closes #565 

**Type of update**
Is this a:
* Bug fix
*  feature enhancement

### DIFF
--- a/plantcv/plantcv/hyperspectral/read_data.py
+++ b/plantcv/plantcv/hyperspectral/read_data.py
@@ -166,10 +166,9 @@ def read_data(filename):
     max_pixel = float(np.amax(array_data))
     min_pixel = float(np.amin(array_data))
 
-    try:
-        wavelength_units = header_dict["wavelength units"]
-    except:
-        wavelength_units = "nm"
+    wavelength_units = header_dict.get("wavelength units")
+    if wavelength_units is None:
+        wavelength units = "nm"
 
         # Create an instance of the spectral_data class
     spectral_array = Spectral_data(array_data=array_data,
--- a/plantcv/plantcv/hyperspectral/read_data.py
+++ b/plantcv/plantcv/hyperspectral/read_data.py
@@ -164,7 +164,12 @@ def read_data(filename):
     max_pixel = float(np.amax(array_data))
     min_pixel = float(np.amin(array_data))
 
-    # Create an instance of the spectral_data class
+    try:
+        wavelength_units = header_dict["wavelength units"]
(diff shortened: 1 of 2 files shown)


plantcv/plantcv/hyperspectral/read_data.py
```diff
@@ -107,7 +107,9 @@ def read_data(filename):
     # Initialize dictionary
     header_dict = {}
 
-    headername = filename + ".hdr"
+    # Remove any file extension and set .hdr filename
+    filename_base = filename.split(".")[0]
```
nfahlgren, line 111:
```suggestion
    filename_base = os.path.splitext(filename)[0]
```

This code will be a little more robust in case someone has a file name with more than one "." in it.

plantcv/plantcv/hyperspectral/read_data.py
```diff
@@ -164,15 +166,20 @@ def read_data(filename):
     max_pixel = float(np.amax(array_data))
     min_pixel = float(np.amin(array_data))
 
-    # Create an instance of the spectral_data class
+    try:
+        wavelength_units = header_dict["wavelength units"]
+    except:
+        wavelength_units = "nm"
```
nfahlgren, line 172:
```suggestion
    wavelength_units = header_dict.get("wavelength units")
    if wavelength_units is None:
        wavelength units = "nm"
```

This isn't effectively different, but all lines should be tested based on your changes to the test data.

PR #582, opened 2020-06-05
 Spatial Clustering Algorithm

SethPolydore:
Contains the actual script and a grayscale mask to test against.  

***Purpose of Pull request***
Adds a new function called spatial_clustering.
This algorithm automatically counts the number of features in a picture.  These features can be whole plants, flowers, leaves, etc.  It uses a binary mask as an input. 

***Type of Update***
New Feature


plantcv/plantcv/spatial_clustering.py
```diff
@@ -0,0 +1,85 @@
...
+    colors = color_palette(n_clusters_ + 1)
+    dict_of_colors = {}
+    sub_mask = []
+    h, w = backtorgb.shape[:2]
+    image = np.zeros((h, w, 3), np.uint8)
+    for y in range(-1, n_clusters_ + 1):
+        dict_of_colors[str(y)] = colors[y]
+    for y in range(0, n_clusters_):
+        sub_mask.append(np.zeros((h, w, 3), np.uint8))
+    dict_of_colors[str(-1)] = (255, 255, 255)
```
nfahlgren, line 64:
```suggestion
```

By removing -1 from the range above this line can be removed.

plantcv/plantcv/spatial_clustering.py
```diff
@@ -0,0 +1,85 @@
...
+    for z in range(0, len(db.labels_)):
+        if not db.labels_[z] == -1:
```
nfahlgren, line 66:
```suggestion
```

As above, removing the -1 index means this line can be removed.

plantcv/plantcv/spatial_clustering.py
```diff
@@ -0,0 +1,85 @@
...
+    n_clusters_ = len(set(db.labels_)) - (1 if -1 in db.labels_ else 0)
+    colors = color_palette(n_clusters_ + 1)
+    dict_of_colors = {}
+    sub_mask = []
+    h, w = backtorgb.shape[:2]
+    image = np.zeros((h, w, 3), np.uint8)
+    for y in range(-1, n_clusters_ + 1):
+        dict_of_colors[str(y)] = colors[y]
+    for y in range(0, n_clusters_):
+        sub_mask.append(np.zeros((h, w, 3), np.uint8))
```
nfahlgren, line 63:
```suggestion
        sub_mask.append(np.zeros((h, w), np.uint8))
```

For the list of masks I think let's make them binary images (I removed the color dimension).

plantcv/plantcv/spatial_clustering.py
```diff
@@ -0,0 +1,85 @@
...
+    dict_of_colors[str(-1)] = (255, 255, 255)
+    for z in range(0, len(db.labels_)):
+        if not db.labels_[z] == -1:
+            sub_mask[db.labels_[z]][zipped[z][0], zipped[z][1]] = (dict_of_colors[str(db.labels_[z])][2],
+                                                                   dict_of_colors[str(db.labels_[z])][1],
+                                                                   dict_of_colors[str(db.labels_[z])][0])
```
nfahlgren, line 69:
```suggestion
            sub_mask[db.labels_[z]][zipped[z][0], zipped[z][1]] = 255
```

To make the masks binary we just need to set the value of each pixel to 255.

plantcv/plantcv/spatial_clustering.py
```diff
@@ -0,0 +1,85 @@
...
+
+        image[zipped[z][0], zipped[z][1]] = (dict_of_colors[str(db.labels_[z])][2],
+                                             dict_of_colors[str(db.labels_[z])][1],
+                                             dict_of_colors[str(db.labels_[z])][0])
+    params.device += 1
+
+    if params.debug == 'print':
+        print_image(image, "full_image_mask.png")
+        for c in range(0, len(sub_mask)):
+            print_image(sub_mask[c], "subimage_cluster_" + str(c) + ".png")
```
nfahlgren, line 79:
```suggestion
```

I think we can print only the colorized image for debugging purposes.

plantcv/plantcv/spatial_clustering.py
```diff
@@ -0,0 +1,85 @@
...
+    elif params.debug == 'plot':
+        plot_image(image)
+        for c in range(0, len(sub_mask)):
+            plot_image(sub_mask[c])
```
nfahlgren, line 83:
```suggestion
```

Same as above, no need to plot the individual masks.

plantcv/plantcv/spatial_clustering.py
nfahlgren: @SethPolydore pointed out we do actually need to account for a -1 label, for points not assigned to a cluster. In the example I was working with there were no unassigned, but it can happen.

PR #592, opened 2020-07-09
Color_clustering_alhorithm

SethPolydore:
contains the scripts for the functionsto automatically segment images by color.  There are two scripts: color_cluster_train (in the learn subpackage) which trains the model that does the segmentation and color_cluster_segmentation which does the actual segmenting.

**Purpose**
Adds new functionalities.


nfahlgren:
Hi @SethPolydore, I suggest refactoring the line causing the error in the `tests.py` file to:

```python
spmask = pcv.color_clustering_segmentation(img=img, project_name=os.path.join(TEST_DATA, "temp_testing_multi"))
```

`os.path.join` will use the appropriate OS-specific file path separators.

HaleySchuhl:
Archiving

PR #593, opened 2020-07-14
Sort width and height to normalize ratio

nfahlgren:
**Describe your changes**
Differences in output from `cv2.minAreaRect` in OpenCV < 3.4.10 and 3.4.10 order width and height differently. We use the ratio of width / height to find contours that are square-ish in `pcv.transform.find_color_card`. To normalize the ratio, this pull request updates `find_color_card` to sort the width and height values and take the ratio of the longest / shortest. Since we do not use the orientation of the bounding rectangle in this function, this meets our purpose of determining which boxes are approximately square.

**Type of update**
Is this a: Bug fix - it will allow us to pass tests with OpenCV >= 3.4, including 3.4.10.


nfahlgren:
@HaleySchuhl, could you take a look at this when you get a chance? I think you had some other ideas about how to get around the issue, feel free to propose alternatives!

nfahlgren:
Haha, sounds good!

Issue #470, opened 2019-10-21
update flags in to be more consistent between a workflow and plantcv-workflow.py


dschneiderch:
New issue for discussion about flags in workflows vs plantcv-workflow.py in parallel package from pull request #457 

I think all the flags are quite confusing. In particular, its confusing that the same letter can mean different things in my workflow vs plantcv-workflow. Anything to standardize them and make them more intuitive would make it easier for people to use.

in VIS workflow:

-i, --image = Input image file
-o, --outdir = output directory
-r, --result = result file
-w, --writeimg = write out image (although you could use this as a T/F flag for anything)
-D, --debug = debug state

in parallel:

-d, --dir = directory of input images
-p, --workflow = workflow to use
-i, --outdir = output directory
-D, --dates = dates to select

so in particular, the d s, i s, and o s
I think it would be more intuitive if they were universal where possible:

-i,--input (this is an image in vis-workflow, a directory for parallel workflow)
-d, --dates = dates to select
-o, --outdir = output directory
-D, --debug = debug state (workflows only)
-O, --debugout = debug_outdir (workflows only)

I will also note that sample_images uses another set of i,o flags:

-s, --source = source or input directory
-o, --outdir = output directory

I would replace -s with -i

finally, yes, I think these would be good (although I was going to ask about the above how you felt about having them be case sensitive.) Personally I like to use the flagnames so I don' t have to remember what each letter means. The benefit of -p, --processes is if you ever want to add -T, --threads then its available. processes is recognizable IMHO.

Workflow script file: -w, --workflow 
Parallel processes: -p, --processes
Save analysis images: -W, --writeimg

last, i find -c, --create a little awkward. at first I thought that it wouldn't make the results file unless i had -c. This is really --overwrite if I understand correctly. -f for force since we already use -o? - but then --meta might have to become -m

I would also like to specify -L (or similar) to specify a parent directory for the timestamped directories that are created with every run of `plantcv-workflow.py`

nfahlgren:
I'm writing up a couple related issues and have a question that we have sort of discussed before. Should we keep developing/refining the input flags to `plantcv-workflow.py` or would it be easier for users (especially on different platforms) if we instead adopted a configuration file-based system for configuring workflow execution? It would be the same kind of inputs but instead of using argparse with the short and long flags we could just have a simple key-value pair file:

```
input       =
workflow    =
metadata    = 
dates       =
outdir      =
debugout    =
processes   = 
imgformat   =
[...]

PR #607, opened 2020-08-16
Config-based parallel processing

nfahlgren:
**Describe your changes**
This pull request reimplements and reworks some ideas started in #502 but limits the changes to a single feature for easier review. In this pull request, a new class is added to the `plantcv.parallel` subpackage that is used to set, manage, and validate configuration options for parallelization of PlantCV workflows. The class `WorkflowConfig` has the following attributes:

```
input_dir         = input images directory (validates that it exists)
json              = output JSON data file (validates that it is provided)
filename_metadata = list of metadata terms filenames are composed of (validates that at least one term is provided)
workflow          = user-defined PlantCV workflow script (validates that it exists)
output_dir        = output directory for images and other outputs (default = ".")
tmp_dir           = an optional parent directory for the automatically created tmp directory (default = None)
processes         = number of parallel processes (default = 1)
start_date        = process images created after this datetime (default = 1)
end_date          = process images created before this datetime (default = None)
imgformat         = image file format/extension (default = "png")
delimiter         = delimiter or regex for parsing metadata terms in filenames (default = "_")
metadata_filters  = dictionary of metadata terms and values used to filter input images (default = None)
timestampformat   = format of image metadata timestamps (default = '%Y-%m-%d %H:%M:%S.%f')
writeimg          = should output images be saved or not (default = False)
other_args        = list of other arguments to pass to workflows (default = [])
```

The class can be used to save or import configuration files, so parameters can either be modified in Python or in an external editor.

Several functions in the `parallel` subpackage were modified to adapt them to using an instance of the `WorkflowConfig` class.

`plantcv-workflow.py` was updated to accept a configuration file as the only input, or alternatively use the legacy command-line inputs for backwards compatibility/preference.

**Type of update**

**Associated issues**
#470 

**Additional context**
These changes will enable us to address some other wish-list items but it will be easier to get these updates tested and integrated first.

**Suggested improvements**
Thanks @dschneiderch!
- [x] In JSON, strings need to be in double, not single quotes - Update PR #609 
- [x] In JSON, regex strings need to have metacharacters escaped (e.g. `"\d"` in Python is `"\\d"` in JSON) - Update PR #609 
- [x] `start_date` and `end_date` should be datetime strings instead of Unix time. Format should match `timestampformat`
- [x] `config.validate_config()` should validate input datetime strings
- [x] Make sure `config.validate_config()` validates metadata terms used in `filename_metadata`
- [x] Add option `cleanup = True/False` to indicate whether `tmp_dir` should be deleted automatically or saved
- [x] Add function to remove `tmp_dir`
- [x] Create `tmp_dir` and `img_outdir` if they do not exist
- [x] Rename `output_dir` parameter to `img_outdir`
- [x] Add parameter `append_results = True/False` to either append to existing output JSON file or overwrite


docs/pipeline_parallel.md
```diff
@@ -8,9 +8,39 @@ to run PlantCV over a flat file directory (just keep this in mind).
...
+```bash
+plantcv-workflow.py --config config.json
+```
+
+### Command-line parameters
+
+Alternative command-line parameters for the `plantcv-workflow.py` script that does the parallelization
 * -d is the --dir directory of images
 * -p is the --workflow that you are going to run over the images, see the [VIS tutorial](vis_tutorial.md) and [PSII tutorial](psII_tutorial.md)
 * -i is the --outdir your desired location for the output images
```
dschneiderch, line 46: -i for --outdir isn't very intuitive

nfahlgren: Agreed, but I didn't change any of the preexisting command-line arguments for backward compatibility. One option moving forward is that we consider the command-line arguments a legacy feature that will eventually be phased out (my current assumption). But alternatively, if desired by folks, we could try to maintain command-line arguments in lock-step with the config-based parameters and have the nomenclature line up

plantcv/parallel/job_builder.py
```diff
@@ -50,64 +39,64 @@ def job_builder(meta, valid_meta, workflow, job_dir, out_dir, coprocess=None, ot
...
+            if meta[img]['imgtype'] != config.coprocess:
                 images.append(img)
         else:
             images.append(img)
-
-    print("Job list will include " + str(len(images)) + " images" + '\n', file=sys.stderr)
+    
+    # Log the number of jobs to be run
+    n_jobs = len(images)
+    print(f"Job list will include {n_jobs} images", file=sys.stderr)
```
dschneiderch, line 51:
```suggestion
try:
    os.makedirs(config.tmp_dir)
except OSError:
    print ("Creation of the directory %s failed" % config.tmp_dir)
```

plantcv/parallel/job_builder.py
```diff
@@ -50,64 +39,64 @@ def job_builder(meta, valid_meta, workflow, job_dir, out_dir, coprocess=None, ot
...
-                     "--outdir", out_dir, "--result", os.path.join(job_dir, img) + ".txt"]
+        job_parts = ["python", config.workflow, "--image", meta[img]['path'],
+                     "--outdir", config.output_dir, "--result",
+                     os.path.join(config.tmp_dir, img) + ".txt"]
         # Add job to list
-        if coprocess is not None and ('coimg' in meta[img]):
-            job_parts = job_parts + ["--coresult", os.path.join(job_dir, meta[img]['coimg']) + ".txt"]
-        if writeimg:
+        if config.coprocess is not None and ('coimg' in meta[img]):
+            job_parts = job_parts + ["--coresult", os.path.join(config.tmp_dir, meta[img]['coimg']) + ".txt"]
```
dschneiderch, line 95: will this fail if config.tmp_dir is not None but doesn't exist?

docs/pipeline_parallel.md
dschneiderch: 👍🏼

plantcv/parallel/job_builder.py
nfahlgren: When tmp_dir is None, the `tempfile` package I am using automatically uses a system temporary directory I believe. But I think you are right, if you specify tmp_dir as a location that does not exist, `tempfile` tries to make a temp directory in that space and will fail.

nfahlgren:
The adaptor parameter is no longer needed, I think (but worth double checking). In the metadata parser I just look for a `SnapshotInfo.csv` file in the input directory. If it is there then I "know" its "Phenofront" format, if it's not I treat it as a flat directory of images.

nfahlgren:
For `create`, I'm open to suggestions. RIght now, if the output JSON file exists, it will append to it. That's sort of both a feature but also a potential issue if one isn't aware that it's doing that. Maybe we could add an option to set the behavior like: `append_results = True/False`?

nfahlgren:
`start_date` and `end_date` are currently pretty ugly, they have to be specified in Unix time. That should really be changed so that datetime strings that match `timestampformat` can be entered. Start/end are currently both inclusive, it rejects images taken < start_date or > end_date.

dschneiderch:
`append_results` seems useful to me.  I always overwrite my results because usually I made a change to the segmentation and now need to rerun everything. But, I'm slowly trying to work towards downloading images every couple days and then running the workflow. Ideally I could set it up to check for new image files (e.g. via Snakemake) in which case I would want to append_results. I think for this to work as well as I'd like I still need to find a near universal color segmentation method, eg like @lsx1980 is doing.

dschneiderch:
I think at some point we discussed saving the timestamp as a global variable in ISO format or something to make it easier to deal with time subsets

nfahlgren:
Make-style-wise, this PR and #609 do not really address that. If you run the pipeline again it will reanalyze everything specified. I think we could implement a simple checkpoint system that keeps track of what was already completed in theory. I did investigate Snakemake as a potential workflow engine but it didn't handle parallelization the way I was looking for. I went with Dask in #609.

dschneiderch:
>     * [ ]  In JSON, strings need to be in double, not single quotes - Update PR #609  
>     * [ ]  In JSON, regex strings need to have metacharacters escaped (e.g. `"\d"` in Python is `"\\d"` in JSON) - Update PR #609

i pushed some doc updates for these two points. I went to the #609 though and it looks like the doc files have some differences. can we merge updates from this pull request into 609?

nfahlgren:
That's a good idea, easier to combine them now and then update.

nfahlgren:
#609 is now merged here

docs/parallel_config.md
```diff
@@ -0,0 +1,114 @@
...
+
+- **Parameters:**
+    - None
+- **Context:**
+    - Used to run validation checks on configuration settings
+
+- **Parameters:**
+    - input_dir (str, required): path/name of input images directory (validates that it exists)
+    - json (str, required): path/name of output JSON data file (appends new data if it already exists)
+    - filename_metadata (list, required): list of metadata terms used to construct filenames
```
HaleySchuhl, line 40: We should have an example of the formatting needed here. This takes a list of strings (in quotes) which should match up with valid metadata terms.

docs/parallel_config.md
```diff
@@ -0,0 +1,114 @@
...
+# Create a template configuration file
+config.save_config(config_file="my_config.json")
+# Edit configuration file as needed
+# Import the configuration file
+config.import_config(config_file="my_config.json")
+# Check for errors
+passed = config.validate_config()
+
+# Change configuration values directly in Python
+config.input_dir = "./my_images"
```
HaleySchuhl, line 113: include example code for how to execute on the command line as well?

dschneiderch:
@nfahlgren re: windows dask test failing, I found this thread. I hvaen't dialed all the way down into the different links but it might help with the issue.
https://github.com/dask/dask/issues/5452

dschneiderch:
did/should #520 get merged into this code base?

nfahlgren:
@dschneiderch an interesting tidbit in that thread is that `setup.py test` is being deprecated: https://github.com/pypa/setuptools/issues/1684. Since that's where the issue is on Windows, we might just bypass it altogether.

nfahlgren:
#520 is a big change that we should look at merging into these changes after we merge this PR into the main branch. This one is already pretty big :)

dschneiderch:
@nfahlgren I thought the whole setup.py thing was basically out of favor with the exception of when you have external code (C, C++, etc) that runs as part of the package

nfahlgren:
I haven't seen what an alternative is. The Python Packaging guide still talks about creating a setup.py file: https://packaging.python.org/tutorials/packaging-projects/

dschneiderch:
hmmm i probably misunderstood or am confusing things.  something about moving away from source distributions and towards wheels
https://flit.readthedocs.io/en/latest/rationale.html and links therein

dschneiderch:
@nfahlgren  curious why you decided converting the date was better in parsers than plantcv-workflow?

nfahlgren:
With this setup we can keep the config properties `start_date` and `end_date` as strings and only convert them to Unix time when we need to but without overwritting the config. Also, currently (until #617), code covered by tests is only in modules, not the stand-alone scripts, so I try to keep `plantcv-workflow.py` as small as possible. This also keeps most of the functionality within the modules that could be reused in principal.

nfahlgren:
Alright, I'm pushing the big green merge button :)

Issue #28, opened 2016-04-29
New function: Add queuing system/batch processing option


nfahlgren:
Develop a queuing system/batch processing option to add on in addition to the current local/multiprocessing capabilities.

nfahlgren:
Will investigate [Parsl](http://parsl-project.org/) as a potential solution to parallelize PlantCV on different environments rather than building a unique solution for different systems.

nfahlgren:
My initial inclination was to utilize an existing workflow engine (e.g. Nextflow, Parsl, Snakemake, etc.). I have tried them all out, at least a bit. I like them a lot but am not sure they work for precisely what I have been trying to achieve. That being said @gsainsbury86 has developed a Nextflow workflow that we need to look at (https://github.com/aus-plant-phenomics-facility/plantcv-pipeline), and Parsl 0.9 (now released) had some planned features I was waiting for, so I should check them out again. Another ideas below...

nfahlgren:
Another way to go I started mapping out was to use `dask`, specifically [dask-jobqueue](https://jobqueue.dask.org/). In doing so, one way we could go would be to eliminate `plantcv-workflow.py` completely. `plantcv-workflow.py` is just a script that parses (extensive) command-line options and then runs functions in the `plantcv.parallel` subpackage.

Right now, a user develops a workflow, likely in Jupyter. If in Jupyter, they convert it to a Python script and have to form it into a workflow script with some argument parsing and plugging their code into the `main()` function. Then they run `plantcv-workflow.py` over their data with that script.

What if we turned this around a bit? Rather than using `plantcv-workflow.py` to execute a workflow script, we just add the parallelization components to the workflow script itself.

Rather than having command-line arguments we could have inputs in a configuration dictionary (though a user could make the config an input easily). We discussed this in #470. Then we package the `plantcv.parallel` functions into a function that runs through the process of parallelizing the workflow.

I would image a user starts with a script downloaded from Jupyter:

```python
from plantcv import plantcv as pcv
img, path, filename = pcv.readimage(image)
```

Then this gets converted (roughly) to this (perhaps automatically with a converter):

```python
from plantcv import plantcv as pcv
from plantcv.parallel import parallelize

def main():
    config = {
        "dir": "./images",
        "json": "pcv2.output.json",
        "outdir": "./output",
        "meta": "imgtype,camera,frame,zoom,lifter,gain,exposure,id",
        "match": "imgtype:VIS,camera:SV,zoom:z1,frame:0",
[...]

PR #609, opened 2020-08-17
Add distributed computing feature

nfahlgren:
**Describe your changes**
This pull request builds on #607 and should not be merged until after it is merged. Here, we remove the older Python multiprocessing parallelization feature and replace it with a configurable [Dask](https://dask.org/) cluster. Python multiprocessing was local parallelization (i.e. computing took place where the program `plantcv-workflow.py` was run). This functionality is maintained by using the `dask.distributed` `LocalCluster` option. However, using `dask-jobqueue`, we now support remote, distributed computing using cluster options available through [dask-jobqueue](https://jobqueue.dask.org/en/latest/), currently: HTCondorCluster, LSFCluster, MoabCluster, OARCluster, PBSCluster, SGECluster, and SLURMCluster.

To configure the parallel computing environment, you create a `WorkflowConfig` object as in #607. The `processes` option has been removed and is replaced with two parameters:

`cluster` = A single string of any of the following `"LocalCluster", "HTCondorCluster", "LSFCluster", "MoabCluster", "OARCluster", "PBSCluster", "SGECluster", and "SLURMCluster"`

`cluster_config` = a dictionary of Dask cluster configuration options. All clusters require `n_workers`. All clusters other than `LocalCluster` also require `cores`, `memory`, and `disk`.

`n_workers` = The size of the cluster you want to use (essentially equivalent to processes previously)
`cores` = The number of processors to use per job. Should be left at 1 in most use cases.
`memory` = The amount of RAM used per job.
`disk` = The amount of disk space needed per job.

There are various other parameters that can be set and advanced configurations used/needed depending on the cluster system you want to use. I added defaults (`None`) for three other common parameters: `log_directory`, `local_directory`, and `job_extra`.

**Type of update**

**Associated issues**
#28 

**Additional context**
To Do
- [ ] Additional testing
- [x] Update documentation

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,20 +5,14 @@ USER root
 
 RUN apt-get update && apt-get install -y libgl1-mesa-glx && rm -rf /var/lib/apt/lists/*
 
-# Install PlantCV
-RUN conda install --quiet --yes -c conda-forge \
-    "matplotlib>=1.5" \
-    "numpy>=1.11" \
-    pandas \
-    python-dateutil \
-    "scipy<1.3" \
-    "scikit-image<0.15" \
-    plotnine \
-    "opencv<4,>=3.4" && \
-    conda clean --all -f -y
-
 # Copy source files
 COPY . /tmp
+
+# Change working directory and modify requirements.txt
+RUN cd /tmp && sed -i'' -e 's/opencv.*//g' requirements.txt
+
+# Install PlantCV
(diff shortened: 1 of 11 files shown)


nfahlgren:
Tested the same dataset on my MacBook and on our HTCondor cluster. Pretty minimal configuration difference to get it going between the two. The only trick in the cluster is getting jobs to work right with conda environments, but overcomeable.

PR #615, opened 2020-09-03
Add overlay_two_imgs to visualize subpackage

DannieSheng:
**Describe your changes**
A function that overlays two images, images can be either RGB or grayscale. 
Along with this function, another function called _resize_img is also in this pull request, which resizes the image to the desired size: crop when larger than the desired size, zero-padding when smaller than the desired size. This function is used in overlay_two_imgs and can also be used in visualize.time_lapse_video, both of the functions require the exact same size of the input images.

It also includes updating features discussed in #634  

**Type of update**
Is this a:
* Update to documentation

**Additional notes**
* Documentation for pcv.transform.resize and pcv.transform.resize_factor to be included


HaleySchuhl (approved):
I made a few minor updates to the documentation page and the internal documentation within the functions. Otherwise it looks great!

DannieSheng:
> I made a few minor updates to the documentation page and the internal documentation within the functions. Otherwise it looks great!

Awesome! Thanks!

nfahlgren:
@DannieSheng I started testing out a service (Codacy) for code review. My intention was to use it as a guide on the backend but it looks like it's enabled itself on the PR checklist even though I don't have it configured that way. So feel free to ignore it for now

DannieSheng:
> @DannieSheng I started testing out a service (Codacy) for code review. My intention was to use it as a guide on the backend but it looks like it's enabled itself on the PR checklist even though I don't have it configured that way. So feel free to ignore it for now

Got it! Thanks!

HaleySchuhl (approved):

nfahlgren:
@DannieSheng I propose removing the image resizing feature of `overlay_two_imgs`, at least for now. I think in most cases, images will be the same size already because in most common workflows one would be overlaying an image with a transformation of the same image (e.g. a binary mask). In cases where the images are not the same, the user could use the `pcv.transform.resize` function separately to resize one or both of the images.

That being said, if later we find that resizing is a common need for the overlay function we could then add the resizing feature back in. Or alternatively, a function that takes in an image series and resizes the whole stack to be the same size could be useful and could be used upstream of the overlay function.

What do you think?

DannieSheng:
@nfahlgren I agree. I think that makes sense.

dschneiderch:
hi all, my experience is that images of different sizes aren't going to overlay particularly well.  I recently did some testing but basically the best results by far for overlaying images from different cameras require a perspective warp. i'm happy to contribute code I put together

nfahlgren:
Thanks @DannieSheng! Do you want to work on that update, or happy to if you are busy with other things.

@dschneiderch, we would definitely be interested in the calibration and data fusion stuff you have been working on!

DannieSheng:
@nfahlgren Yes I'd love to!

DannieSheng:
@dschneiderch Awesome! Good to hear that!

plantcv/plantcv/visualize/overlay_two_imgs.py
```diff
@@ -48,28 +45,12 @@ def overlay_two_imgs(img1, img2, alpha=0.5, size_img=None):
...
-    if size_img is None:
-        size_img = np.max([sz_img1[0], sz_img2[0]]), np.max([sz_img1[1], sz_img2[1]])
+    # assumption: the sizes of img1 and img2 are the same
+    size_img = img1_.shape[0:2]
 
     # initialize the output image
     out_img = np.zeros(size_img + (3,), dtype='uint8')
 
-    # check if sizes are the same
-    if sz_img1 != size_img:
```
dschneiderch: if I may, I think we should check that the sizes are the same. if not, raise a fatal error and ask the user to resize one of the images e.g. with tranform subpackage

nfahlgren: Added, thanks @dschneiderch!

nfahlgren (approved):
Thank you @DannieSheng!

PR #625, opened 2020-09-17
fix plotting bug when debug=None

### LINKED ISSUE #620: Pseudocolor function plotting issues
**Describe the bug**
@dschneiderch noted that on his Windows system the `pcv.visualize.pseudocolor` function was plotting even when `pcv.params.debug = None`. He proposed the updating the debug logic block to the following:

```python
if params.debug is not None:
    if params.debug == 'print':
        plt.savefig(os.path.join(params.debug_outdir, str(params.device) + '_pseudocolored.png'), dpi=params.dpi)
        plt.close()
    elif params.debug == 'plot':
        plot_image(pseudo_img1)
        # Use non-blocking mode in case the function is run more than once
        plt.show(block=False)
```

Essentially this removes an extra piece of logic that did `plt.show()` even when debug was set to `None`.

I think there are some other issues to work out though, so I'm opening this as a new issue instead of applying the fix within another PR.

**To Reproduce**

**Issue 1**
We automatically set the `matplotlib` backend under conditions where we do not detect a display.:

```python
# If there is no display or a matplotlib backend already defined, use the non-GUI backend
if "DISPLAY" not in os.environ and "MPLBACKEND" not in os.environ:
    matplotlib.use("Agg")
```

On Linux and macOS, if `DISPLAY` is not a defined environment variable, or we are not in Jupyter (MPLBACKEND), then we set the backend to Agg so that plots can be created without trying to plot to a window/screen device. One part of @dschneiderch issue I think is that on Windows we are not accounting for display devices.

**Issue 2**
Related to the above issue, `pcv.visualize.pseudocolor`, with the current code, displays the image even when debug is `None`. In Jupyter for example:

```python
%matplotlib inline
pcv.params.debug = None
img, imgpath, imgname = pcv.readimage(filename="image.png")
gray_img = pcv.rgb2gray_hsv(rgb_img=img, channel="v")
ps_img = pcv.visualize.pseudocolor(gray_img=gray_img)
# Plots the pseudocolor image here
ps_img.savefig("test.png")
# Saves the image correctly to a file
```

**Issue 3**
If the current code is modified to remove the extra `plt.show()` and the code is run multiple times in an environment with the `Agg` backend multiple scale bars are plotted, probably because the figure does not get closed before plotting again. 

**Expected behavior**
I'm not 100% sure what the overall solution is, we will need to try running some test code in Jupyter, on the command line, and with/without plotting interfaces to make sure we can get the behavior set to be consistent.

 - OS: macOS, Windows
 - Environment: conda, Jupyter, command-line
 - PlantCV Version: 3.9.0+68.g409b26b

dschneiderch:
**Describe your changes**
I removed the plotting in pseudocolor() when debug=None because it causes issues if the matplotlib backend is not set correctly.  I also removed setting the display in  `__init__.py` because I think matplotlib takes care of this appropriately.  I added a plt.close() for the current figure at the end of pseudocolor() if debug=None.  this prevents the duplicate colorbars when running pseudocolor() multiple times but does NOT prevent the return of the figure handle. see comments in #620 for example.

I tested is on Windows with an ipython terminal, a .py script run with ipython, and using the plantcv workflow method. I also tested with debug='plot' and debug='print' in an ipython terminal on Windows.  Also tested successfully with jupyter notebook 

**Type of update**
Is this a:
* Bug fix 

**Associated issues**
closes #620

### DIFF
--- a/plantcv/plantcv/__init__.py
+++ b/plantcv/plantcv/__init__.py
@@ -1,8 +1,5 @@
 import os
 import matplotlib
-# If there is no display or a matplotlib backend already defined, use the non-GUI backend
-if "DISPLAY" not in os.environ and "MPLBACKEND" not in os.environ:
-    matplotlib.use("Agg")
 
 observations = {}
 
--- a/plantcv/plantcv/plot_image.py
+++ b/plantcv/plantcv/plot_image.py
@@ -23,17 +23,17 @@ def plot_image(img, cmap=None):
         matplotlib.rcParams['figure.dpi'] = params.dpi
         # If the image is color then OpenCV stores it as BGR, we plot it as RGB
         if len(dimensions) == 3:
-            fig = plt.figure()
+            plt.figure()
             plt.imshow(cv2.cvtColor(img, cv2.COLOR_BGR2RGB))
             plt.show()
 
         elif cmap is None and len(dimensions) == 2:
-            fig = plt.figure()
+            plt.figure()
             plt.imshow(img, cmap="gray")
             plt.show()
 
         elif cmap is not None and len(dimensions) == 2:
-            fig = plt.figure()
+            plt.figure()
             plt.imshow(img, cmap=cmap)
             plt.show()
 
--- a/plantcv/plantcv/plot_image.py
+++ b/plantcv/plantcv/plot_image.py
@@ -23,14 +23,17 @@ def plot_image(img, cmap=None):
         matplotlib.rcParams['figure.dpi'] = params.dpi
         # If the image is color then OpenCV stores it as BGR, we plot it as RGB
         if len(dimensions) == 3:
+            fig = plt.figure()
             plt.imshow(cv2.cvtColor(img, cv2.COLOR_BGR2RGB))
             plt.show()
 
         elif cmap is None and len(dimensions) == 2:
+            fig = plt.figure()
             plt.imshow(img, cmap="gray")
             plt.show()
 
         elif cmap is not None and len(dimensions) == 2:
+            fig = plt.figure()
             plt.imshow(img, cmap=cmap)
             plt.show()
 
--- a/plantcv/plantcv/threshold/threshold_methods.py
+++ b/plantcv/plantcv/threshold/threshold_methods.py
@@ -245,10 +245,11 @@ def triangle(gray_img, max_value, object_type="light", xstep=1):
     # Additional figures created by this method, if debug is on
     if params.debug is not None:
         if params.debug == 'print':
-            plt.plot(hist)
-            plt.title('Threshold value = {t}'.format(t=autothreshval))
-            plt.axis([0, 256, 0, max(hist)])
-            plt.grid(True)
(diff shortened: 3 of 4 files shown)


dschneiderch:
The other question is whether other functions need the same fix? doesn' tlook like any others do anything when debug=None but could they run into similar issues like the duplicate scalebar?

nfahlgren:
We have talked recently about some other plotting changes. I'm going to add them here but it would be great if you could make sure they work on your end too. Want to make sure it's an improvement on all platforms.

nfahlgren:
By creating a new figure each time we run `plot_image` it fixes an issue when using `%matplotlib notebook` mode where it used to send all plots to the open Figure 1. Now it creates a new plot each time without closing the existing plots.

dschneiderch:
seems to work here. every figure is in tis own window thingy even when run in the same cell in a notebook.
```
%matplotlib notebook
img,_,_ = pcv.readimage('tests/data/input_color_img.jpg')
img2,_,_ = pcv.readimage('tests/data/VIS_SV_0_z300_h1_g0_e85_v500_93054.png')
pcv.plot_image(img)
pcv.plot_image(img2)
```

nfahlgren:
Added the figure thing in a couple more places where plots were done separately, including `pseudocolor`.

nfahlgren (approved):

dschneiderch (commented):
my brain is tired but quick thoughts.
1. does plt.figure need to be assigned to fig?  fig is not used. we just need to create a new figure handle for the notebook backend to work properly?
2. same question for pseudo_img1 and pseudo_img2. they are not used so plt.imshow can just be printed on the current figure handle and then you get the current figure handle at the end with gcf. 

maybe you already tried these. other wise we can test tomorrow.

nfahlgren:
Not sure. I followed a matplotlib tutorial and that's how they did it. I agree it's weird to have unused variables. To me it's actually kind of an annoying thing about matplotlib, that creating a figure in this case changed the behavior of downstream steps without explicitly passing variables. Sort of like base R graphics

nfahlgren:
I have seen code like `_, ax = plt.subplots()` though, so probably don't technically need the fig variable or others.

dschneiderch:
> 
> 
> I have seen code like `_, ax = plt.subplots()` though, so probably don't technically need the fig variable or others.

oh yeah, that would probably do it.

there are a couple places in plantcv in general where you guys use function()[0] instead of x,_=function() and its very confusing to read :-)

dschneiderch:
> 
> 
> Not sure. I followed a matplotlib tutorial and that's how they did it. I agree it's weird to have unused variables. To me it's actually kind of an annoying thing about matplotlib, that creating a figure in this case changed the behavior of downstream steps without explicitly passing variables. Sort of like base R graphics

base R? who uses those?

dschneiderch:
awesome though, glad to see this finally getting addressed !

nfahlgren:
Removing the unused variables did not seem to break anything :)

PR #637, opened 2020-10-27
force unix line endings except for any windows specific files

dschneiderch:
**Describe your changes**
use .gitattributes to force unix line endings so windows git doesnt think there were a million changes
exclude windows specific files like .cmd and .bat

https://code.visualstudio.com/docs/remote/troubleshooting#_resolving-git-line-ending-issues-in-containers-resulting-in-many-modified-files

**Additional context**
windows woes


nfahlgren (approved):
Awesome, thanks @dschneiderch!

PR #645, opened 2020-11-16
update dockerfile to conda-forge

dschneiderch:
The vice dockerfile is still refering to bioconda instead of conda-forge for pcv > 3.8 (i think thats what it was)

I'm looking at these because I was wondering if we could tag the docker images with the version# instead of 'latest'  It looks like a new docker image is created for every merge into master. is that correct? I have no idea how it gets uploaded though.  However for the sake of creating reproducible environments for analyses it would make more sense to tag the image files with the released version number.  can you make a 'latest' alias to the latest version?

### DIFF
--- a/vice/Dockerfile
+++ b/vice/Dockerfile
@@ -15,7 +15,7 @@ USER jovyan
 # install plantcv
 RUN conda update -n base conda \
     && conda install jupyterlab=0.35.4 nb_conda \
-    && conda create -n plantcv -c bioconda plantcv nb_conda \
+    && conda create -n plantcv -c conda-forge plantcv nb_conda \
     && conda clean -tipsy \
     && fix-permissions $CONDA_DIR
 


nfahlgren:
The main Docker image (https://hub.docker.com/r/danforthcenter/plantcv) is updated every time we merge into the main branch because we build PlantCV from source. The only tag we have defined is latest, though any merge into the main branch can be used I think (https://hub.docker.com/r/danforthcenter/plantcv/builds). 

For VICE, the container installs PlantCV from conda so only official versioned releases would ever have new container images, though they are still built with every merge.

nfahlgren:
We have the GitHub repo hooked up to Docker Hub using a webhook. I think if we wanted to tag containers with releases we would potentially need to do it through the GitHub Actions `.github/publish-package.yml` workflow

nfahlgren:
Potentially with this Action: https://github.com/marketplace/actions/build-and-push-docker-images

dschneiderch:
> 
> 
> We have the GitHub repo hooked up to Docker Hub using a webhook. I think if we wanted to tag containers with releases we would potentially need to do it through the GitHub Actions `.github/publish-package.yml` workflow

this might be an option
https://medium.com/better-programming/build-your-docker-images-automatically-when-you-push-on-github-18e80ece76af

HaleySchuhl:
no longer have docker function

PR #646, opened 2020-11-17
Minor morphology updates 

### LINKED ISSUE #654: Debug has no effect when line thickness set in canny_edge_detect
**Describe the bug**
In the function `pcv.canny_edge_detect` when the input `thickness` is not equal to 1 (default) then `pcv.params.debug` = "print" or "plot" will have no effect due to if/else statement.

**To Reproduce**
Current function logic:

```python
    # Adjust line thickness
    if thickness != 1:
        debug = params.debug
        params.debug = None
        bin_img = dilate(bin_img, thickness, 1)
        params.debug = debug
    else:
        # Print or plot the binary image
        if params.debug == 'print':
            print_image(bin_img, os.path.join(params.debug_outdir, (str(params.device) + '_canny_edge_detect.png')))
        elif params.debug == 'plot':
            plot_image(bin_img, cmap='gray')
```

**Expected behavior**

I think the intended logic is:

```python
    # Adjust line thickness
    if thickness != 1:
        debug = params.debug
        params.debug = None
        bin_img = dilate(bin_img, thickness, 1)
        params.debug = debug
    
    # Print or plot the binary image
    if params.debug == 'print':
        print_image(bin_img, os.path.join(params.debug_outdir, (str(params.device) + '_canny_edge_detect.png')))
    elif params.debug == 'plot':
        plot_image(bin_img, cmap='gray')
```

HaleySchuhl:
**Describe your changes**
A handful of minor changes to functions inside the morphology sub-package. 1. segment insertion angle was failing on skeleton structures that should have been successful, so a tweak to the part of the algorithm that combines the stem segments into a single object helps. 2. branch and tip points should be integers. 3. bugfix within `pcv.canny_edge_detect` function where the debug parameter had no effect when the line thickeness is set. 

**Type of update**
Is this a:
* Bug fix

**Associated issues**
* closes #654 

**Additional context**


plantcv/plantcv/morphology/segment_insertion_angle.py
```diff
@@ -164,6 +162,8 @@ def segment_insertion_angle(skel_img, segmented_img, leaf_objects, stem_objects,
         intersection_angles.append(intersection_angle)
 
     # Compile list of measurements where there is a 'NA' where pruned away segments would go.
+    print(intersection_angles)
+    print(pruned_away)
```
nfahlgren, line 166: Are these debugging print statements that can be removed?

nfahlgren (approved):

Issue #634, opened 2020-10-16
Refactor and improve image resizing


nfahlgren:
**Is your feature request related to a problem? Please describe.**
The function `plantcv.resize` could be improved in a few ways. It currently allows for resizing the x and y dimensions using scaling factors, but resizing to an exact size may be desirable also. We also use a single, fixed interpolation method that may not yield optimal results in all cases.

**Describe the solution you'd like**
I propose a few modifications to `plantcv.resize` and a new function.

1. Move `plantcv.resize` to the `transform` subpackage: `plantcv.transform.resize`. This is both because it is a transformation of the input image and it's consistent with `scikit-image`.
2. Refactor `plantcv.transform.resize` to resize images to an exact size.
3. In `plantcv.transform.resize`, if the new size is smaller than the input size, use the OpenCV interpolation method `cv2.INTER_AREA`. If the new size is larger than the input size, use the interpolation method `cv2.INTER_CUBIC`. (These are the generally recommended options, but we could point a user to the OpenCV function if they want more control than what we provide).
4. Make a new function called something like `plantcv.transform.resize_factor` that combines the new features proposed for `plantcv.transform.resize` but uses the current feature of scaling the sizes by factors rather than exact sizes.
5. To avoid redundancy, it may be useful to have a helper function set the interpolation method.

**Describe alternatives you've considered**
These functions could be wrapped into one like the OpenCV function, but I think it's simpler to have two smaller functions rather than twice the inputs to one.

DannieSheng:
Not sure if it is related, I will also be interested in adding resizing functions that resize images to exact sizes, either by cropping or padding. Currently, I have this function defined inside the function generating time_lapse_video.

nfahlgren:
Hi @DannieSheng, it did come to my mind while I was reviewing your PR! I realized that the resize function you proposed there does not use interpolation though, it either crops or pads as you said. Maybe now that I think about it, these could potentially be one function?

```
def resize(img, size, interpolation=True):
    # img = numpy.ndarray
    # size = tuple
    # interpolation = bool

    img_area = img.shape[0] * img.shape[1]
    new_area = size[0] * size[1]

    if interpolation:
        if new_area >= img_area:
            resized = cv2.resize(img, dsize=size, interpolation=cv2.INTER_CUBIC)
        else:
            resized = cv2.resize(img, dsize=size, interpolation=cv2.INTER_AREA)
    else:
        if new_area >= img_area:
            # pad
        else:
            # crop
[...]

PR #648, opened 2020-11-18
Perspective warp

dschneiderch:
**Describe your changes**
Add function to fuse images that have different size/scale/proportions, e.g. from two (different) cameras in different locations . The function accepts 2 images and 4 pairs of points in order to compute a perspective transformation matrix so the images overlay nicely. I found this works much better than simple scale/rotate/translate https://cougphenomics.github.io/camera_overlays.html

**Type of update**
Is this a:

**Associated issues**
discussed in various forms here #615 and #634


dschneiderch:
@nfahlgren @HaleySchuhl  The tests are failing on colorspaces() function which I didn't touch !?!  It looks like it isn't importing the rgb2gray_* functions properly but i have no idea why.

nfahlgren:
@dschneiderch I think I got it, it was a tricky one. The `transform` submodule is imported before the `visualize` submodule because `visualize.colorspaces` uses `transform.resize_factor`. The new `transform.warp` function now uses `visulize.overlay_two_imgs`. That's actually fine, but `visualize.colorspaces` also uses `rgb2gray_hsv` and `rgb2gray_lab`, which are imported after `transform`. So since `transform` imports from `visualize` I just had to import `rgb2gray` functions before `transform`. Clear as mud right? :)

dschneiderch:
wow, well thanks. I would not have guessed that import order matters. that means there could be circular dependencies(!)

nfahlgren:
@dschneiderch this is great! I just tested it on one of our old use cases, fitting a mask derived from an RGB image onto an image from our NIR cabinet. One little issue I ran into that I need to think about more is that our NIR images are 16-bit and the mask is 8-bit, so when they get overlayed in the debug image it looks like weird noise. I'm not sure that it needs to be fixed, I just converted the NIR image to 8-bit to get around it.

In thinking about this function and our existing `crop_position_mask` I'm wondering if we should consider deprecating `crop_position_mask` in favor of `warp`? I think this new approach is easier to use and is more likely to produce a good fit because it takes into account more than just scaling and translation. I can imagine a scenario where it's not possible to pick landmark points and `crop_position_mask` could still be used, but not sure that is going to be a common situation. What do @dschneiderch and @HaleySchuhl think?

dschneiderch:
Thanks!
My first thought is to internally convert `refimg` to 8bit for the display but that might be problematic if you have really low values - for example from a bioluminescence image. Maybe we should convert `img` to the type of `refimg`? It'd be a little more user friendly.
From a theoretical standpoint `warp` is just a more advanced geometric transformation than scale and translate used in `crop_position_translate` (which I think could be vastly simplified by switching to opencv functions). What scenario were you thiking about not picking landmark points? You should be able to get the same results with `warp` as with `crop_position_mask`. if you know your cropping parameters `x` and `y` than you can use them to warp too. simplest case, the `refimg` coords are simply the image limits. I'd suggest labeling `crop_position_mask` as soft deprecated and point people towards `warp`. Hard for me to imagine a case when `crop_position_mask` is more useful than `warp` - maybe just easier to wrap your brain around what's happening.

nfahlgren:
Yeah, just imagining a case where there's nothing in the background to use as anchor points. With `crop_position_mask` you could guess and check until you got the overlap the way you want but I think that would be harder with landmark points. But like I said, not sure that's a likely scenario and I do think `warp` is not only a more complete computational approach, but also easier if you have easy to identify landmark points (i.e. no guess and check necessary).

dschneiderch:
You can always use the edges and center of the plant as landmarks and adjust from there, no?

nfahlgren:
True, that makes sense. So maybe we can mark `crop_position_mask` with a deprecation warning for the remainder of v3 and remove it in v4, if folks agree with that plan.

HaleySchuhl:
That sounds like a good plan to me!

nfahlgren (approved):
Thanks @dschneiderch, this is a nice function to have integrated in!

PR #649, opened 2020-11-19
correct readbayer behavior

### LINKED ISSUE #642: readbayer color channels are out of order
**Describe the bug**
readbayer() is producing color images where the bands are out of order and image is rotated.  I am actually getting a RGB image back (with bayertype 'BG') which means all the pcv functions are converting it to BGR when displaying, for example.

I tracked down `cvtColor` bug that seems to exist on and off in opencv from 2.4 -> 4.1 https://github.com/opencv/opencv/issues/4857
The last version on conda forge with the *correct* behavior is 3.4.2.  (did not try 4.x series)

**To Reproduce**
download https://github.com/danforthcenter/plantcv/files/4314758/Pats.Wheat.Root.Growth_002-0001_A1_2019-06-11_07-19-08.461_0.zip

```
import cv2
filename='data/images/Pats Wheat Root Growth_002-0001_A1_2019-06-11_07-19-08.461_0.tif'
image_raw=cv2.imread(filename, -1)
img = cv2.cvtColor(image_raw, cv2.COLOR_BayerBG2BGR)
pcv.plot_image(img)
```

image

**Expected behavior**

image
or
even better rotated too which happens with opencv 3.4.2
image

 windows 10, conda, pcv 3.10.1 and latest master. seems dependent on opencv >3.4.2

**Additional context**
2 possible solutions until upstream gets fixed:

1. limit opencv to 3.4.2

2. switch the color rendering to color demosaicing package

dschneiderch:
it was discovered that readbayer() resulted in a different band order with more recent opencv than the one installed when the function was written. turns out opencv at the time the function was written was giving an incorrect band order so this makes sure that installations avoid problematic opencv versions and changes the default bayer pattern to be what it was meant to be

**Describe your changes**
bump opencv min version to 3.4.7 to avoid bayer pattern bug in 3.4.2 and 3.4..4. new default bayerpattern is RG to match original (false) visual behavior

**Type of update**
Is this a:
* Bug fix

**Associated issues**
closes #642


HaleySchuhl:
Archiving for now

Issue #662, opened 2021-01-05
Add optional "label" parameter for all analyze_* functions 


HaleySchuhl:
Prefix the observation variable label (since that is the key to the dictionary and must be unique, this gets around the issue with the over writing if the user tries to use the same analyze function within one workflow) by adding a parameter. 

**Describe the solution you'd like**

**Additional context**
- from discussion #660

PR #663, opened 2021-01-07
Add "label" parameter to morph&hyperspectral analysis fxns

HaleySchuhl:
**Describe your changes**
Add an optional `label` parameter to all functions that store data to the Outputs class so that we can allow for repeated measures and more descriptive data labels. 

**Type of update**
Is this a:
* Work in progress

**Associated issues**
#662 
#660 

**Additional context**


nfahlgren:
@HaleySchuhl what do you think about simplifying the code a bit so no extra tests would be required for the label parameter?

We could set a default label (maybe `label="default"` ?). Then because label would have a value by default you could remove the:

```python
if label == None:
    prefix = ""
else:
    prefix = label + "_"
```

And just replace it with `prefix = label + "_"`

HaleySchuhl:
> @HaleySchuhl what do you think about simplifying the code a bit so no extra tests would be required for the label parameter?

That's a great idea! I can make that adjustment in both PRs

HaleySchuhl:
I'm currently looking at the `Outputs` doc page, and I came across a question. @nfahlgren is it true that the addition of a `label` parameter for all analysis functions makes the `args.coresult` functionality redundant? Before it let people save the shape of the RGB and NIR data, but now they could just run the function twice (or more) per workflow without clearing and saving new output files.

nfahlgren:
@HaleySchuhl good question. It doesn't currently make it redundant but is something that will help us deprecate the coresult system eventually. The label parameter would allow us to save NIR results as part of the same output file as the VIS results, but coresult also functions in `plantcv-workflow.py` as a mechanism to organize the VIS and NIR inputs to the workflow. Once we have a working solution to #423 we won't need coresult anymore.

PR #664, opened 2021-01-08
fix duplicate on workflow metadata image (path+name+name)

JorgeGtz:
**Describe your changes**
The workflow parallelization was duplicating the name of the image for the metadata "image". Removed the concatenation of the path (including name) and the name. 

**Type of update**
Is this a bug fix


nfahlgren (approved):
Thanks @JorgeGtz!

PR #678, opened 2021-01-27
Add an attribute "include_all_subdirs" to WorkflowConfig

DannieSheng:
Current plantcv parallel workflow would go though all subdirectories in the given input_dir and process all images whose metadata meet the criteria. By adding this "include_all_subdirs" attribute to WorkflowConfig class, users have the option to include all subdirectories by setting "include_all_subdirs" to `True` (by default it is True) or only include images in the input_dir by setting "include_all_subdirs" to `False`.

Is this a:
* Update to documentation

A solution to discussion #677


docs/parallel_config.md
```diff
@@ -45,6 +45,8 @@ Validate parameters/structure of configuration data.
 
 **workflow**: (str, required): path/name of user-defined PlantCV workflow Python script (validates that it exists).
 
+**include_all_subdirs**: (bool, default = `True`): not including images in subdirectories if `False`
```
nfahlgren, line 48:
```suggestion
**include_all_subdirs**: (bool, default = `True`): If `False`, only images directly in `input_dir` (no subdirectories) will be analyzed.
```

nfahlgren (changes requested):
Looks great @DannieSheng and I tested it out with `plantcv-workflow.py` locally and it worked perfectly! I'm just suggesting a small edit to the documentation description.

nfahlgren (approved):

Issue #667, opened 2021-01-14
Please support the newest version of opencv


JimXu1989:
Thanks a lot for your sharing !!! Great work!!!

while when I install plantcv, it install old version of opencv by default, I want use the newest version of opencv, 4.5.1, what should I do?

nfahlgren:
Hi @JimXu1989, apologies for the delay! Our plan is to migrate to OpenCV when we start the PlantCV v4 series in the next few months. I tested it out on the current version 3.10.1 and it almost works, but we need to update line 302 in `plantcv/plantcv/roi/roi_methods.py` to make it OpenCV version agnostic. That line currently expects 3 outputs, but in OpenCV4 there are only 2. We only care about those two anyhow, so that will be an easy fix that we can do now. Once that's done you should be able to use OpenCV4 even though it wouldn't officially be in our requirements file yet

JimXu1989:
> Hi @JimXu1989, apologies for the delay! Our plan is to migrate to OpenCV when we start the PlantCV v4 series in the next few months. I tested it out on the current version 3.10.1 and it almost works, but we need to update line 302 in `plantcv/plantcv/roi/roi_methods.py` to make it OpenCV version agnostic. That line currently expects 3 outputs, but in OpenCV4 there are only 2. We only care about those two anyhow, so that will be an easy fix that we can do now. Once that's done you should be able to use OpenCV4 even though it wouldn't officially be in our requirements file yet

plantcv's fill region method really better then opencv's crossponding method !!

PR #682, opened 2021-02-05
Update OpenCV function call for cross-version compatibility

nfahlgren:
**Describe your changes**
Updates the `pcv.roi.multi` function to modify the `cv2.findContours` function call to only keep the last two output items (the contours and hierarchy). The current code expects a third output before these that only exists in OpenCV 3 but not version 4. Since the first output was not used, it can be discarded for cross-compatibility between 3 and 4.

**Type of update**
Is this a:
* Minor update

**Associated issues**
#667


DannieSheng (approved):
I agree with this change!

Issue #692, opened 2021-02-22
apply_mask() got an unexpected keyword argument 'rgb_img'


Gustavofh:
Hello, I'm having trouble using the apply_mask() method. 

When I run the script provided in vis_tutorial on my computer, an error is generated when applying this method. This is not recognizing the 'rgb_img' argument, only 'img'. Moreover, in documentation, rgb_img argumnt don't appear.

I am using the latest version of plantcv (3.11.0).

How should I proceed? 

below follows code and message error:

```
# Appy Mask (for VIS images, mask_color='white')

#   Inputs:
#   rgb_img - RGB image data 
#   mask - Binary mask image data 
#   mask_color - 'white' or 'black' 
masked = pcv.apply_mask(rgb_img=img, mask=bs, mask_color='white')

---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-12-9652b53c479e> in <module>
      5 #   mask - Binary mask image data
      6 #   mask_color - 'white' or 'black'
----> 7 masked = pcv.apply_mask(rgb_img=img, mask=bs, mask_color='white')

TypeError: apply_mask() got an unexpected keyword argument 'rgb_img'
```

I believe there is an error in relation to the version presented in the tutorial (analyzed on the cyverse platform) with the one I have on my machine. However, the script runs perfectly on the jupyter notebook open directly from the cyverse.

 - Linux
 - Conda environment
 - PlantCV Version [3.11.0]

nfahlgren:
Hi @Gustavofh,

I thought we had fixed this once before but it looks like the VIS tutorial documentation page needs to be updated. The correct input to `pcv.apply_mask` is `img` and not `rgb_img` starting with PlantCV v3.7. The Python script code at the bottom of the VIS tutorial page is correct, but the code above still has some instances of the incorrect argument `rgb_img`.

Thanks for pointing this out, we will update the documentation as soon as possible.

HaleySchuhl:
@GrantKonkel I found three other instances that needs updating with this same oversight when we updated the `pcv.apply_mask` function to make it more generic. Two are in the VIS/NIR tutorial, and one in the script at the bottom of the hyperspectral tutorial. 

I will go through today and push updates to the interactive tutorials hosted on binder (I'll push commits to the PlantCV binder [repo](https://github.com/danforthcenter/plantcv-binder) to correct anything I find that needs updating there! But when you get a chance could you take care of these 3 lines from the PlantCV docs?

[...]

PR #696, opened 2021-02-23
Fix syntax error in apply_mask

GrantKonkel:
**Describe your changes**
Updated the vis-tutorial and deleted "rgb_" from the "img".

**Type of update**
Is this a:
* Bug fix

**Associated issues**
#692


nfahlgren (approved):
Thanks @GrantKonkel!

Issue #657, opened 2020-12-07
Move rotate function to transform submodule


nfahlgren:
Move `pcv.rotate` to `pcv.transform.rotate`:

From https://github.com/danforthcenter/plantcv/pull/615#issuecomment-722592435

> I think it would make sense to move rotate to `transform.rotate()` (maybe keep an alias in pcv.rotate -> pcv.transform.rotate())

nfahlgren:
Completed in #697

PR #697, opened 2021-02-23
Relocate rotate

GrantKonkel:
**Describe your changes**
Moved the rotate function to the transform sub-package
Created a pcv.rotate placeholder so the rotate function is backwards compatible

**Type of update**
Is this a:

**Associated issues**
#657

### DIFF
--- a/docs/rotate2.md
+++ b/docs/rotate2.md
@@ -2,7 +2,7 @@
 
 Rotates image without changing the dimensions of the image.
 
-**rotate**(*img, rotation_deg, crop*)
+**plantcv.transform.rotate**(*img, rotation_deg, crop*)
 
 **returns** image after rotation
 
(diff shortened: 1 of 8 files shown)


docs/rotate2.md
```diff
@@ -28,7 +28,7 @@ from plantcv import plantcv as pcv
 pcv.params.debug = "print"
 
 # Rotate image
-device, rotate_img = pcv.rotate(img, 10, True)
+device, rotate_img = pcv.transform.rotate(img, 10, True)
```
nfahlgren, line 31:
```suggestion
rotate_img = pcv.transform.rotate(img, 10, True)
```

nfahlgren (changes requested):
Just one small change and this PR should be good to merge. We had an old variable `device` still in the `rotate` documentation that should be removed.

nfahlgren (approved):

PR #712, opened 2021-03-07
Create debug module

nfahlgren:
**Describe your changes**
Adds a private function `_debug` that handles the decision-making to save or display visuals for debugging steps. The if/elif blocks in each PlantCV function that determine whether to use `pcv.print_image` or `pcv.plot_image` can be removed and replaced with a call to `_debug`:

```python
from plantcv.plantcv import params
from plantcv.plantcv._debug import _debug
_debug(visual=img, filename=os.path.join(params.debug_outdir, str(params.device) + "_debug_img.png")
```

Tests for each PlantCV function will no longer need to run each function with debug = "print" or "plot" enabled.

**Type of update**

**Associated issues**
#706, #713

### DIFF
--- a/docs/CONTRIBUTING.md
+++ b/docs/CONTRIBUTING.md
@@ -167,19 +167,25 @@ import cv2
 import os
 
 # Import functions from within plantcv
-from plantcv.plantcv import params 
-from plantcv.plantcv import plot_image
-from plantcv.plantcv import print_image
+from plantcv.plantcv import params
+from plantcv.plantcv._debug import _debug
 ```
 
 Here is a sample of a new function. Arguments should be defined.
-Generally, functions should utilize the global `debug` parameter from the [params](params.md) class.
-Users can set `pcv.params.debug `
+Generally, functions should utilize the private function `_debug` to produce visualization results.
+Users can set `pcv.params.debug`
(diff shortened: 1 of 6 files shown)


HaleySchuhl (approved):

HaleySchuhl:
@nfahlgren One consideration with consolidating testing is that we have different types of images output. Maybe we could test one numpy array visual and one matplotlib figure visual in the tests?

DannieSheng:
I just found that there is no     
```
# Auto-increment the device counter
params.device += 1
```
in this `readimage` function as other PlantCV functions do. So that when in "print" debug mode, the saving name would be the same when reading multiple input images.

nfahlgren:
@DannieSheng how about this, what if we add the device counter code:

```python
# Auto-increment the device counter
params.device += 1
```

to the `_debug` function and remove it from all the other functions as part of the update we are working on for #713?

DannieSheng:
@nfahlgren I think it makes sense to me.

nfahlgren:
We could have a pull request to add this code to `_debug` for now. It will cause the device counter to get over counted until we remove the extra counters from other functions, but I don't think that will hurt anything in the interim.

DannieSheng:
I can include this change in my pull request #734 (which has pretty small changes and ready to review), or do we want to open a new pull request?

DannieSheng:
Yes it makes sense to do so!

Issue #700, opened 2021-02-25
Expand functionality of `pcv.cluster_contours_splitimg` 


HaleySchuhl:
**Describe the bug**
The function `pcv.cluster_contours_splitimg` has an `rgn_img` parameter, but since the function was added we have added the support for more image types. This method of multiplant analysis is totally amenable to other datatypes, so I propose that we change the parameter name to a more generic `img` and update the source code such that grayscale/thermal/fluor data can also get processed in this way. 

**To Reproduce**
- [ ] update `rgb_img` to `img` 
- [ ] test the updated function on a grayscale image 
- [ ] update cluster_contours_splitimg.md
- [ ] update updating.md
- [ ] update multiplant tutorial doc page 
- [ ] update instances in tests/tests.py (find all instances of pcv.cluster_contours_splitimg) 
- [ ] 

**Additional context**

PR #719, opened 2021-03-09
Pcv.cluster contours splitimg accept non rgb images

GrantKonkel:
**Describe your changes**
Previously the splitimg() function did not take in non RGB images, which was changed.

**Type of update**
Is this a:
* Update to documentation

**Associated issues**
#700

### DIFF
--- a/plantcv/plantcv/cluster_contour_splitimg.py
+++ b/plantcv/plantcv/cluster_contour_splitimg.py
@@ -9,7 +9,7 @@ from plantcv.plantcv import apply_mask
 from plantcv.plantcv import params
 
 
-def cluster_contour_splitimg(rgb_img, grouped_contour_indexes, contours, hierarchy, outdir=None, file=None,
+def cluster_contour_splitimg(img, grouped_contour_indexes, contours, hierarchy, outdir=None, file=None,
                              filenames=None):
 
     """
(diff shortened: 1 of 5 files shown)


plantcv/plantcv/cluster_contour_splitimg.py
```diff
@@ -119,9 +119,9 @@ def cluster_contour_splitimg(rgb_img, grouped_contour_indexes, contours, hierarc
         else:
             savename = os.path.join(".", group_names[y])
             savename1 = os.path.join(".", group_names1[y])
-        iy, ix, iz = np.shape(rgb_img)
+        iy, ix, iz = np.shape(img)
```
nfahlgren, line 122:
```suggestion
        iy, ix = np.shape(img)[:2]
```

It looks like we never use the variable `iz` so we can just use a slice to return only the first two dimensions, then it won't matter if `img` is RGB or grayscale.

nfahlgren (changes requested):
This looks great @GrantKonkel!

I added one direct suggestion here on GitHub that you can take a look at and accept as a commit if you think it makes sense to change.

The one other suggestion I have is that we might want to add a second test function in `tests.py` for `cluster_contour_splitimg` where the input image is grayscale since the updated function will support it. The new test does not need to be as big as the current test and can use the same input data. But in the current test where we have the input image as `img1 = cv2.imread(os.path.join(TEST_DATA, TEST_INPUT_MULTI), -1)`, in the grayscale test we can just change that line to `img1 = cv2.imread(os.path.join(TEST_DATA, TEST_INPUT_MULTI), 0)`, where the zero flag will read the image as grayscale. The new test could just run `cluster_contour_splitimg` once with `pcv.params.debug = None`.

nfahlgren (approved):

Issue #713, opened 2021-03-07
Refactor functions to utilize a debug module


nfahlgren:
**Is your feature request related to a problem? Please describe.**
Pull request #712 creates a new private function `_debug` to replace the existing debug image display logic in most PlantCV functions. The new function has a single test function with two parameters, "print" and "plot". By utilizing the `_debug` function in existing PlantCV functions, a logical if/elif block can be removed from each and replaced with a single call to `_debug`. Tests for those functions can also be refactored to drop tests with "print" and "plot" enabled, which will reduce testing run time and complexity.

**Describe the solution you'd like**
PR #712, when merged will add the `_debug` private function without interfering with existing functions. It also refactors `pcv.readimage`, `pcv.analyze_object`, and `pcv.analyze_color` for testing purposes. I propose that, once #712 is merged, we open a new pull request per PlantCV function as time allows to update the rest of the code base over time. I recognize that is a lot of pull requests, but it will be quick to review and merge them, versus trying to update hundreds of files and tests in a single PR. If you are working on an existing function for other reasons, feel free to include these updates in that PR though.

For a given function, the general framework involves updating the function by removing imports to `print_image` and `plot_image` and adding an import for `_debug`. Then removing the logic around any debugging visuals and passing them, along with a filename to `_debug`. So for an existing pattern like:

```python
import os
from plantcv.plantcv import params
from plantcv.plantcv import print_image
from plantcv.plantcv import plot_image

# function code here

if params.debug is not None:
    # make a debug image
    if params.debug == "print":
        print_image(debug_img, os.path.join(params.debug_outdir, str(params.device) + "_function.png"))
    elif params.debug == "plot":
        plot_image(debug_img)
```

It can be simplified to:

```python
import os
from plantcv.plantcv import params
from plantcv.plantcv._debug import _debug

# function code here
# make a debug image
_debug(visual=debug_img, filename=os.path.join(params.debug_outdir, str(params.device) + "_function.png"))
```

Likewise, an existing test pattern like this:

```python
def test_plantcv_function():
    # Test cache directory
    cache_dir = os.path.join(TEST_TMPDIR, "test_plantcv_function")
    os.mkdir(cache_dir)
    pcv.params.debug_outdir = cache_dir
    # Test with debug = "print"
    pcv.params.debug = "print"
    _ = pcv.function()
    # Test with debug = "plot"
    pcv.params.debug = "plot"
    _ = pcv.function()
    # Test with debug = None
[...]

PR #720, opened 2021-03-10
Update warp functions

DannieSheng:
**Describe your changes**
Extend previous version of pcv.transform.warp such that not only binary or grayscale images are acceptable input images. 
Change the function name to "warp_align" because this function calculates the transformation from target image to reference image based on user provided landmark points, then warp the target to make it aligned to the reference image. 

**Type of update**
Is this a:
* Feature enhancement and new feature 
* Update to documentation
* Work in progress

**Associated issues**
#717 

**Additional context**
Will also update the documentation when the interactive framework of collecting landmark points becomes available.
Also finishes pcv.transform.warp in #713


DannieSheng:
There can be more optional parameters for robust estimation functions, i.g. RANSAC, LMED, and RHO.
https://docs.opencv.org/master/d9/d0c/group__calib3d.html#ga4abc2ece9fab9398f2e560d53c8c9780
https://docs.opencv.org/3.0-beta/doc/py_tutorials/py_feature2d/py_feature_homography/py_feature_homography.html 

However, if including those parameters, the function would be more complicated.

plantcv/plantcv/transform/warp.py
```diff
@@ -1,58 +1,82 @@
...
+from plantcv.plantcv import params
+from plantcv.plantcv._debug import _debug
 from plantcv.plantcv import fatal_error
 from plantcv.plantcv import color_palette
-from plantcv.plantcv import params
 from plantcv.plantcv.visualize import overlay_two_imgs
+from plantcv.plantcv.transform import rescale
+from skimage import img_as_ubyte
 
+def _preprocess_img_dtype(img):
```
dschneiderch, line 16: do we need this function?? I think you can just send img through `img_as_ubyte()` and it will return what we need.  also rescale doesn't work on rgb images which are supported in  the rest of the function. one caveat is that the scikit-image function will only accept float between -1,1 and will clip negative numbers upon conversion to uint8. I think we can live with that though. most float data would be between 0-1 for images(?) and mostly all plantcv functions only work with uint8

plantcv/plantcv/transform/warp.py
```diff
@@ -1,58 +1,82 @@
...
+    :return mat: numpy.ndarray
     :return warped_img: numpy.ndarray
+    :return mat: numpy.ndarray
     """
 
-    params.device += 1
+    if len(pts) != len(refpts):
+        fatal_error('Please provide same number of corresponding coordinates.')
 
-    if len(pts) != 4 or len(refpts) != 4:
```
dschneiderch, line 36: this function will fail if the number of points given is less than 4 so we should check for that

plantcv/plantcv/transform/warp.py
```diff
@@ -1,58 +1,82 @@
...
+    rows_img, cols_img = shape_img[0:2]
 
     # convert list of tuples to array for cv2 functions
     ptsarr = np.array(pts, dtype='float32')
     refptsarr = np.array(refpts, dtype='float32')
 
     # find tranformation matrix and warp
-    mat, _ = cv2.findHomography(ptsarr, refptsarr, method=methods.get(method))
+    mat, status = cv2.findHomography(ptsarr, refptsarr, method=methods.get(method))
+    if mat is None:
```
dschneiderch, line 78: good catch!

plantcv/plantcv/transform/warp.py
dschneiderch: or proceed img_as_ubyte() with 
if img.dtype == 'uint8':

plantcv/plantcv/transform/warp.py
```diff
@@ -61,51 +85,99 @@ def warp(img, refimg, pts, refpts, method='default'):
...
-            cv2.drawMarker(img2,
-                           pt,
-                           color=colors[i],
-                           markerType=cv2.MARKER_CROSS,
-                           markerSize=params.marker_size*res_ratio_i,
-                           thickness=params.line_thickness*res_ratio_i)
-
-        # temp rgb image for colored markers on refimg
-        refimg2 = refimg.copy()
+            if status[i][0] == 1:
```
dschneiderch, line 109: nice feature!  just add it to the documentation

plantcv/plantcv/transform/warp.py
```diff
@@ -61,51 +85,99 @@ def warp(img, refimg, pts, refpts, method='default'):
...
+
+
+def warp_align(img, mat, refimg=None):
+    """
+    Warp the input image based on given transformation matrix mat, to align with the refimg
+
+    Inputs:
+    img = image to be warped
+    mat = transformation matrix (size: (3,3))
+    refimg = (option) reference image. default=None
```
dschneiderch, line 154: the output of this function doesn't make much sense if refimg isn't set here. 
image
it seems reasonable to me to require the output image size if warping an image.

plantcv/plantcv/transform/warp.py
```diff
@@ -61,51 +85,99 @@ def warp(img, refimg, pts, refpts, method='default'):
...
-            print_image(refimg2, os.path.join(params.debug_outdir, str(params.device) + "_img-ref.png"))
-            print_image(img_blend, os.path.join(params.debug_outdir, str(params.device) + "_warp_overlay.png"))
+        _debug(visual=img_marked, filename=os.path.join(params.debug_outdir, str(params.device) + "_img-to-warp.png"))
+        _debug(visual=refimg_marked, filename=os.path.join(params.debug_outdir, str(params.device) + "_img-ref.png"))
+        _debug(visual=img_blend, filename=os.path.join(params.debug_outdir, str(params.device) + "_warp_overlay.png"))
+
+    return warped_img, mat
+
+
+def warp_align(img, mat, refimg=None):
```
dschneiderch, line 147: i like that you can create and reuse the transformation matrix!

plantcv/plantcv/transform/warp.py
```diff
@@ -1,58 +1,82 @@
 # warp image
 
 import cv2
+import skimage
```
dschneiderch, line 4: dont need this

plantcv/plantcv/transform/warp.py
```diff
@@ -1,58 +1,82 @@
...
 import os
 import numpy as np
-from plantcv.plantcv import print_image
-from plantcv.plantcv import plot_image
+import copy
```
dschneiderch, line 7: dont think you need this either

tests/tests.py
```diff
@@ -977,6 +977,12 @@ def test_plantcv_outputs_add_observation_invalid_type():
         outputs.add_observation(sample='default', variable='test', trait='test variable', method='type', scale='none',
                                 datatype=list, value=np.array([2]), label=[])
 
+def test_plantcv_transform_warp_warp_align():
```
dschneiderch, line 980: not sure I understand what this is testing....?

tests/tests.py
```diff
@@ -1038,66 +1044,42 @@ def test_plantcv_transform_warp_smaller():
...
-                                    refpts = [(0,0),(vcol-1,0),(vcol-1,vrow-1),(0,vrow-1)])
-
-    assert np.sum(mask_warped_print)==83103814
+    pcv.params.debug = None
+    warped_img,  mat = pcv.transform.warp(img, img, pts = [(0,0),(vcol-1,0),(vcol-1,vrow-1),(0,vrow-1)], refpts = [(0,0),(vcol-1,0),(vcol-1,vrow-1),(0,vrow-1)])
+    assert mat.shape == (3, 3) and warped_img.dtype == img.dtype
 
+    _ = pcv.transform.warp_align(img, mat)
+    warped_img_ = pcv.transform.warp_align(img, mat, img)
+    assert warped_img_.dtype == img.dtype
```
dschneiderch, line 1055: based on what @nfahlgren  said yesterday we can't have multiple assert per test because assert acts like return.  also look like this checks the dtype same as before

tests/tests.py
```diff
@@ -1038,66 +1044,42 @@ def test_plantcv_transform_warp_smaller():
...
+    # different number of points
+    # pcv.params.debug = None
     with pytest.raises(RuntimeError):
-        _ = pcv.transform.warp(img, img,
-                                pts = [(0,0),(mcol-1,0),(mcol-1,mrow-1),(0,mrow-1)],
-                                refpts = [(0,0),(vcol-1,0),(vcol-1,vrow-1),(0,vrow-1)])
-
+        pcv.transform.warp(img, img, pts=[(0, 0), (vcol-1, 0), (vcol-1, vrow-1),(0,vrow-1)], refpts=[(0, 0), (vcol-1, 0), (vcol-1, vrow-1)])
 
-def test_plantcv_transform_warp_4ptserror():
```
dschneiderch, line 1082: i think we need this test

tests/tests.py
```diff
@@ -1038,66 +1044,42 @@ def test_plantcv_transform_warp_smaller():
...
-    assert np.sum(mask_warped_print)==83103814
+    pcv.params.debug = None
+    warped_img,  mat = pcv.transform.warp(img, img, pts = [(0,0),(vcol-1,0),(vcol-1,vrow-1),(0,vrow-1)], refpts = [(0,0),(vcol-1,0),(vcol-1,vrow-1),(0,vrow-1)])
+    assert mat.shape == (3, 3) and warped_img.dtype == img.dtype
 
+    _ = pcv.transform.warp_align(img, mat)
+    warped_img_ = pcv.transform.warp_align(img, mat, img)
+    assert warped_img_.dtype == img.dtype
 
-def test_plantcv_transform_warp_rgbimgerror():
```
dschneiderch, line 1069: dont need the rgbimg error test anymore but we should test behavior for different number of points

tests/tests.py
```diff
@@ -1038,66 +1044,42 @@ def test_plantcv_transform_warp_smaller():
...
-    pcv.params.debug = None
-    mask_warped = pcv.transform.warp(bimg_small, img[:,:,2],
-                                    pts = [(0,0),(mcol-1,0),(mcol-1,mrow-1),(0,mrow-1)],
-                                    refpts = [(0,0),(vcol-1,0),(vcol-1,vrow-1),(0,vrow-1)])
-    pcv.params.debug = 'plot'
-    mask_warped_plot = pcv.transform.warp(bimg_small, img[:,:,2],
-                                pts = [(0,0),(mcol-1,0),(mcol-1,mrow-1),(0,mrow-1)],
-                                refpts = [(0,0),(vcol-1,0),(vcol-1,vrow-1),(0,vrow-1)])
-
-    assert np.count_nonzero(mask_warped)==93142
```
dschneiderch, line 1051: good to get rid of the tests with precomputed answers per yesterday's conversation. I think this function would lend itself well to parametrize the tests  and we test all combinations of inputs wrt rgb and binary since they could have slightly different behaviors in the future due to # dimensions.

dschneiderch (changes requested):
It works and i didn't manage to break it with my inputs! I did leave some considerations for you though.

I will go through the documentation more after another revision but I left some comments. 

Also looks like tests will need to be reconfigured since they were lumped all into one.

lastly, did this make it into the code?
`plantcv.transform.ImageRegistrator(img_ref, img_tar, figsize=(12, 6))` if not can we delete the doc file?

plantcv/plantcv/transform/warp.py
DannieSheng: The values for thermal images won't be between -1,1, so `pcv.transform.rescale` is used to convert the datatype. To make this function more general, use `skimage.img_as_ubyte` to catch `RuntimeError`. The assumption here is that if an image is a multi-channel one, values are between -1,1.

tests/tests.py
DannieSheng: I've parameterized the tests for `transform.warp` and `transform.warp_align`

tests/tests.py
```diff
@@ -6008,6 +5937,51 @@ def test_plantcv_threshold_texture():
...
+
+    if refimg_ is not None:
+        refimg_ = refimg
+        warped_img_ = pcv.transform.warp_align(img, mat, refimg_)
+        assert warped_img.shape[0:2] == refimg.shape[0:2] and warped_img_.shape[0:2] == refimg_.shape[0:2]
+    else:
+        warped_img_ = pcv.transform.warp_align(img, mat)
+        assert warped_img.shape[0:2] == refimg.shape[0:2] and warped_img_.shape[0:2] == img.shape[0:2]
+
+@pytest.mark.parametrize("img, refimg, pts, refpts, method", [
```
dschneiderch, line 5972: - [x ] fail on different length pts
- [x ] fail on same length but <4
- [ ] fail because can't calc homography?

tests/tests.py
```diff
@@ -6008,6 +5937,51 @@ def test_plantcv_threshold_texture():
...
+def create_test_img(sz_img):
+    img = np.random.randint(np.prod(sz_img), size=sz_img) * 255
+    img = img.astype(np.uint8)
+    return img
+
+def create_test_img_bin(sz_img):
+    img = np.zeros(sz_img)
+    img[3:7, 2:8] = 1
+    return img
+@pytest.mark.parametrize("img, refimg, method, refimg_, pts, refpts, debug_mode", [
```
dschneiderch, line 5949: do we need to test that the debug images work as expected? there is an above average amount of logic in that code. here we at least make sure it runs without an error but that doesn't insure it runs correctly....

tests/tests.py
dschneiderch: maybe we can annotate the different parameterizations for future reference?

plantcv/plantcv/transform/warp.py
```diff
@@ -3,56 +3,87 @@
...
     # convert list of tuples to array for cv2 functions
     ptsarr = np.array(pts, dtype='float32')
     refptsarr = np.array(refpts, dtype='float32')
 
     # find tranformation matrix and warp
-    mat, _ = cv2.findHomography(ptsarr, refptsarr, method=methods.get(method))
+    mat, status = cv2.findHomography(ptsarr, refptsarr, method=methods.get(method))
+    if mat is None:
+        fatal_error("Cannot calculate a robust model with given corresponding coordinates and with desired robust estimation algorithm {}!".format(method))
     warped_img = cv2.warpPerspective(src=img, M=mat, dsize=(cols_ref, rows_ref))
```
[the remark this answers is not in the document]
DannieSheng: I am not sure about this. 
 @nfahlgren I remember you mentioned that we should avoid using plantcv function in test, but what if in this case?

tests/tests.py
DannieSheng: I agree.
Yes, the 3rd failed because of not being able to converge.

plantcv/plantcv/transform/warp.py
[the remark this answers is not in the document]
nfahlgren: I agree generally about avoiding duplication but I think it's a bit simpler, in this case, to run the OpenCV function directly since there's no other pre-processing going on. If we run `warp_align`, we have to cache the debug state, turn it off, turn it back on after, and run through the preprocessing code in `warp_algin` that has no effect here (I think?).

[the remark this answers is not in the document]
DannieSheng: You are right!

plantcv/plantcv/transform/warp.py
nfahlgren: Right, I do think we need to change this somehow. The size of the output image is determined by the `refimg`, so if one is not provided then the transformed image is output at its original size. While that works it does not produce an image that is likely usable in many cases (assuming that typically `img` and `refimg` are different sizes).

One option is to require `refimg`. Another is to require the output shape and not have a `refimg`, which is only used to get the shape and plot an overlay (not needed because fit was already determined by calculating the matrix earlier?).

dschneiderch: it comes down to which one is more annoying:

shape is required and you have refimg:
`refimg.shape`
or
refimg is required and you don't have refimg:
`np.zeros_like((x,y))`

I would assume the user has refimg, for example an image from another camera you are matching.
i'm flexible though.

nfahlgren: True, plus if desired someone could do `img=img, refimg=img`

DannieSheng: This `warp_align` function is there for the cases where the transformation matrix is available, such that we can warp the `img` with or without `ref_img`. The `ref_img` is only used for visualization - the blended image. The returned variable is the `warped_img`

nfahlgren: I was thinking the function of `warp_align` was more that you don't need to recalculate `mat` for each pair of images, you would just need to create it once with a pair of training images. But then on each use, you would still have the pair of `img` and `refimg` (e.g. thermal + RGB, RGB + NIR). 

When you create `mat` based on images of different sizes, if you run `warp_align` without `refimg` you will get an image of size `img` instead of `refimg`, so then it likely won't work for what you want to do with it downstream.

But in the event one really doesn't have a `refimg`, either `img` can be input as both `img` and `refimg` or a blank image using `np.zeros` can be input as `refimg`.

DannieSheng: That's correct. For most of the prospective user cases, both `img` and `refimg` are expected inputs. 
Oh, I see the point, 
`warped_img = cv2.warpPerspective(src=img, M=mat, dsize=(cols_ref, rows_ref))`

The size of the warped image depends on `mat` as well as provided `(cols_ref, rows_ref))`

nfahlgren (approved):
Thanks @DannieSheng, this is really great! And thanks for the thorough review @dschneiderch!

nfahlgren:
@dschneiderch looks like you need to approve the changes, then we can merge. Thanks!

dschneiderch (approved):

PR #726, opened 2021-03-18
add_deprecation_update_histogram_related

DannieSheng:
Describe your changes
Add a deprecation_warning function

Update the pcv.visualize.histogram function such that it is capable of showing histogram not only for grayscale images, but also multi-spectral input images. 

Updated histogram related functions: **pcv.analyze_color**, **pcv.analyze_thermal_values**, **pcv.analyze_nir_intensity**  by
(1)Apply updated version of **pcv.visualize.histogram** function 
(2)Prints deprecation warnings and current version of PlantCV

Type of update
Is this a:

New feature or feature enhancement
Update to documentation
Work in progress

Associated issues
Discussion #685


nfahlgren:
I just started testing this and one thing I observed so far is that you can get a weird histogram if you don't set the lower_bound and upper_bound and the image data don't extend the full range expected. I ran into this with a grayscale image where the min/max values were not 0 and 255, respectively. So when I set bins=256 the function set the range to 74-151 and divided that range by 256, which caused it to plot sparse data.

Without setting the bounds:
Screen Shot 2021-03-22 at 8 41 26 PM

With bounds set
Screen Shot 2021-03-22 at 8 41 37 PM

One option is to leave it as is and have the user add bounds when needed. The other would be to override bins if the number of bins is greater than the distance between the min and max value. The downside to the latter is that it could create inconsistencies when run on different images if the user didn't realize the number of bins and/or the bounds were an issue when testing on sample data.

DannieSheng:
@nfahlgren 

I agree. I remembered I was trying to set default value for `bins` to be the difference between the `upper_bound and` the `lower_bound`, however that would only work for images with `int` values, i.e. won't be ideal for plotting histogram for thermal images.  As shown in the attached examples
So I think it might be better to leave it and let user set the upper and lower bounds?

hist_therm_100
hist_therm_200

nfahlgren:
@DannieSheng I made some changes to the PR and tried to describe them in the commit messages. Please double-check that everything still works the way you expect.

One thing that is probably not working as we might expect is the `deprecation_warning` function. When using the `warnings` package, Python filters out DeprecationWarning and other warning types by default unless you run Python with the `-W` option or use `warnings` to turn the filter off. So right now, in Jupyter, in the Python console, or in a PlantCV workflow script no deprecation warnings will be printed. When you run `pytest` it will print warnings because they are viewed as something only developers want to see. More information about this here: https://stackoverflow.com/questions/20960110/how-do-i-get-warnings-warn-to-issue-a-warning-and-not-ignore-the-line.

I think since we are trying to prompt users to change their workflow syntax we will probably just need to print a message to STDERR instead of using `warnings`. If that makes sense we could update the function:

```python
import sys
from plantcv.plantcv import _version


def deprecation_warning(warning):
    """Print out deprecation warning

    Inputs:
    warning = warning message text

    :param warning: str
    """
    version = _version.get_versions()
    warning_msg = f"DeprecationWarning: {warning} Current PlantCV version: {version['version']} released on {version['date']}"
    print(warning_msg, file=sys.stderr)
```

nfahlgren (approved):
Thanks @DannieSheng!

PR #730, opened 2021-03-21
Bug sample images subfolders

JorgeGtz:
**Describe your changes**
img_element_array now stores the relative path of the images (root+file). That is all that is needed for the part that copies the images.

**Type of update**
Is this a:
* Bug fix

**Associated issues**
The function was not working properly if the image database contained sub-folders


nfahlgren (approved):
Thanks @JorgeGtz!

Issue #747, opened 2021-04-12
Metadata parser check date range does not work with timestamps that only contain time as expected


nfahlgren:
**Describe the bug**
The `plantcv.parallel.metadata_parser` module can parse timestamp metadata from filenames given a timestamp format string specified in `WorkflowConfig` (e.g. `"%Y-%m-%d"`). Users also have the option to specify a date range within which images should be analyzed (`start_date` and `end_date` set in `WorkflowConfig`). When `start_date` and `end_date` are set to `None` (default) then `plantcv.parallel.metadata_parser` sets the start and end date range from 1970-1-1 0:0:1 (epoch) to “now”. But we made an assumption that timestamps would include year, month, and day. In the case when only hour, minute, and/or second is set, the `start_date` and `end_date` end up getting truncated to 00:00 and the current time, so the images selected for analysis are likely not what is expected.

The temporary fix (I think) until we fix this bug is to set `start_date` and `end_date` in the config to `"00-00-00"` and `"23-59-59"` (with an appropriate delimiter for your data).

**Expected behavior**
We need to update `plantcv.parallel.metadata_parser` to handle time-only timestamps.

**Local environment (please complete the following information):**
 - OS: All
 - Environment All
 - PlantCV Version: 3.11

**Additional context**
H/T: Thanks to Briony Parker for identifying this bug.

PR #748, opened 2021-04-13
handle subdaily timestamps

dschneiderch:
**Describe your changes**
checked to see if end_date is less than now. if end_date only includes hours then now will have year 1900 and we thus set end_date to 23-59-59 (when timestampformat is '%H-%M-%S'). see bug report below for issue.
 
a benefit of the proposed solution is that it should work with months and days too if there are no years because I set end_date to 2099-12-31-23-59-59 before formatting with timestampformat. is 2099 long enough!? I tested for the subdaily case only.

**Type of update**
Is this a:
* Bug fix

**Associated issues**
#747 

**Additional context**
test added.


dschneiderch:
i of course totally messed up the merge/commits 🤦‍♂️ but I think the PR is correct!

nfahlgren:
So you're saying PlantCV won't make it past 2099?? :)

Thanks for working on this!

nfahlgren (approved):
Thanks @dschneiderch!

PR #750, opened 2021-04-13
Update crop reporter & Add NPQ

HaleySchuhl:
**Describe your changes**
Update the method for reading in FLUOR data with `pcv.photosynthesis.read_cropreporter`. Now utilizes x-array data format and returned a single data object rather than multiple frames. Also updated `pcv.photosynthesis.analyze_fvfm` to accept the new x-array data structure, and updated to the new `_debug` function method. Added `pcv.photosynthesis.analyze_npq` by copying the logic from the other function. 

**Type of update**
Is this a:
* Work in progress
    *  [x] update fluor_fvfm.md
    *  [x] update read_cropreporter.md
    *  [x] update updating.md
    *  [x] write doc page for analyze_npq
    *  [x] update tests for two existing functions... *in progress*
    *  [x] add test data and write tests for analyze_npq
    *  [ ] add a helper function and/or documentation for how to create labeled xarrays from multiple frames (for users with non-crop reporter formatted data) 

**Associated issues**
- #708 
- 
**Additional context**


dschneiderch:
I think it would be better to stick with xarray methods than pulling out the frames and using numpy directly. 
YII and NPQ can be computed multiple times per dataset so as it is written the user would have to call analyze_npq() for each timestep in the dataset. Fv/Fm is just a special case of YII so in #426 I proposed a generic analyze_yii() function with a `parameter` flag in which you'd specify dark-adapted yii (=fv/fm) vs light-adapted yii. 

for the xarray object you can specify a dimension that defines the pairs of frames of the dataset and then use `groupby` to calculate yii and npq for every point on that dimension.


to compute yii:
```
def calc_yii(da):
    return (da.sel(frame='Fmp') - da.sel(frame='Fp'))/da.sel(frame='Fmp')

yii = da_concat.groupby('parameter', squeeze=True).map(calc_yii).drop_vars('frame')
```

to compute npq:

```
def calc_npq(Fmp, Fm):
    return(Fm/Fmp - 1)

npq = xr.apply_ufunc(calc_npq, da_concat.sel(frame='Fmp').groupby('parameter', squeeze=True), da_concat.sel(parameter='t0',frame='Fmp')).drop_vars('frame')
```

HaleySchuhl:
> I think it would be better to stick with xarray methods than pulling out the frames and using numpy directly.
> YII and NPQ can be computed multiple times per dataset so as it is written the user would have to call analyze_npq() for each timestep in the dataset. Fv/Fm is just a special case of YII so in #426 I proposed a generic analyze_yii() function with a `parameter` flag in which you'd specify dark-adapted yii (=fv/fm) vs light-adapted yii.
> 
> for the xarray object you can specify a dimension that defines the pairs of frames of the dataset and then use `groupby` to calculate yii and npq for every point on that dimension.
> 
> to compute yii:
> 
> ```
> def calc_yii(da):
>     return (da.sel(frame='Fmp') - da.sel(frame='Fp'))/da.sel(frame='Fmp')
> 
> yii = da_concat.groupby('parameter', squeeze=True).map(calc_yii).drop_vars('frame')
> ```
> 
> to compute npq:
> 
> ```
> def calc_npq(Fmp, Fm):
>     return(Fm/Fmp - 1)
> 
> npq = xr.apply_ufunc(calc_npq, da_concat.sel(frame='Fmp').groupby('parameter', squeeze=True), da_concat.sel(parameter='t0',frame='Fmp')).drop_vars('frame')
> ```

I agree! I wanted to get the functions working as expected by recycling existing code first, but I plan to integrate the xarray methods to streamline those functions. I also agree that now would be a good time to change the function name, seeing as this update will come with breaking changes.

nfahlgren:
There's one task listed above missing here: a helper function for making Xarray DataArrays for non-CropReporter data. I propose we add that in a separate PR. Fdark QC analysis was also removed in this PR, I propose we bring in the function @dschneiderch outlined in #426 (a separate QC function rather than built in to the yii function) in a separate PR

plantcv/plantcv/photosynthesis/analyze_yii.py
```diff
@@ -0,0 +1,165 @@
...
+    # Prime is empty for Fv/Fm (dark- and light-adapted) and ' for Fq'/Fm'
+    prime = ""
+    if measurement.lower() == "fq'/fm'":
+        prime = "'"
+    # Create a list of fluorescence values, measurement labels, and timepoint indices
+    fluor_values = []
+    meas = []
+    timepts = []
+    # Iterate over the PAM measurement frames
+    for i in range(ps.attrs[f"F{prime}-frames"]):
```
dschneiderch, line 153: this would be more idiomatic
```
from sklearn.utils import img_as_bool
fluor_values = psii.where(np.expand_dims(img_as_bool(mask),2)).mean(axis=(0,1))
```
and 
`meas = psii.coords['frame_label']`

and then subset for F{\d+}

plantcv/plantcv/photosynthesis/analyze_yii.py
```diff
@@ -0,0 +1,165 @@
...
+
+    # Analyze fluorescence induction curve to identify max fluorescence
+    ind_df = _fluor_induction(ps=ps, mask=mask, measurement=measurement)
+
+    # Make fluorescence induction curve figure using plotnine
+    ind_fig = (ggplot(data=ind_df, mapping=aes(x="Timepoints", y="Fluorescence", color="Measurement"))
+               + geom_line(show_legend=True))
+
+    # Select the F0, F0', or F' frame
+    fmin = ps.sel(frame_label=ps.attrs[prot_vars[measurement.lower()]["F0"]]).data
```
dschneiderch, line 56: if ps has another dimension `parameter` then we can do something like 
```
def calc_yii(da):
    return (da.sel(frame='Fmp') - da.sel(frame='Fp'))/da.sel(frame='Fmp')

yii = ps.groupby('parameter', squeeze=True).map(calc_yii).drop_vars('frame_label')
```
i think we still need to map each F0,... to a Fm' and F' with a dict. maybe that can be an input to calc_yii

plantcv/plantcv/photosynthesis/analyze_yii.py
dschneiderch: but in the end i still had to loop to subset the fluor_values and meas labels
```
idx=[]
for i, s in enumerate(ps.coords['frame_label'].data):
    if re.search(r'\d+', s):
        idx.append(i)

idx        
fluor_values = psii.where(np.expand_dims(img_as_bool(mask),2)).mean(axis=(0,1))[idx]

meas = psii.coords['frame_label'][idx]

```

dschneiderch: oh! they have string methods. 
```
idx = [psii.coords['frame_label'].str.contains('\d+')]
meas = psii.coords['frame_label'][idx]
fluor_values = psii.where(np.expand_dims(img_as_bool(mask),2)).mean(axis=(0,1))[idx]

File: plantcv/plantcv/photosynthesis/analyze_yii.py (hunk 2 above)
Comment (dschneiderch) [reply plantcv/plantcv/photosynthesis/analyze_yii.py:56]:
```
ps_expanded = ps.expand_dims(dim = 'measurement_num', axis=3)
def calc_yii(da, fp, fmp):
    return (da.sel(frame_label=fmp) - da.sel(frame_label=fp))/da.sel(frame_label=fmp)
fvfm = ps_expanded.groupby('measurement_num', squeeze=True).map(calc_yii,args = ('F0', 'F5')).drop('frame_label')
```
replace F0 and F5 with the dict inferno

dschneiderch: woops, i double checked this and it doesn't give what I expected. will need to investigate another day

dschneiderch: assign new coord to existing dimension
http://xarray.pydata.org/en/stable/generated/xarray.DataArray.assign_coords.html?highlight=add%20coords

nfahlgren:
Closing since the functionality was alternately added in #790

Issue #749, opened 2021-04-13
Plant segmentation from Fmax vs Fmin images from CropReporter


annacasto:
I've noticed in my Fv/Fm histograms from CropReporter images that there's a peak around 1.0, which is an unusual value for Fv/Fm. In many cases, this peak is larger than the rest of the histogram, so all the fvfm_hist_peak values in the output data are ~1.0. 
![image](https://user-images.githubusercontent.com/61164490/114579219-44721080-9c43-11eb-97e7-b9bd22828b09.png)

It was unclear from the Fv/Fm pseudocolored images which pixels had these high Fv/Fm values. We determined that in the Fmax images pixels around the border of the plant had large Fmax values and small Fmin values, making Fv/Fm close to 1. 
![image](https://user-images.githubusercontent.com/61164490/114583387-28706e00-9c47-11eb-9bbd-c3fd7c083d06.png)
_Image: Fmax image pseudocolored by pixel values. There's a little halo around the border of the plant._

![image](https://user-images.githubusercontent.com/61164490/114583681-78e7cb80-9c47-11eb-9cf2-3ef16ec366e5.png)
_Image: Fmin image pseudocolored pixel values. Pixel values around the border of the plant are very low._

![image](https://user-images.githubusercontent.com/61164490/114582332-1b06b400-9c46-11eb-9e84-e259b74b29c6.png)
_Image: Fv/Fm values between 0.98 - 1_

In my opinion, this border isn't actually part of the plant. It's more pronounced in side view images, which use a mirror to reflect light to and from the camera, than in top view images. It could also be a leaf movement problem. 

The halo around the plant isn't present in the Fmin images, and segmenting the plant based on Fmin eliminates the Fv/Fm peak around 1. Using the erode function, on the the mask also fixes this issue, but it also ends up masking more of the plant compared to the Fmin-based mask. 

I'm proposing editing the documentation page to recommend segmenting based on the Fmin image instead of Fmax, or adding an erosion step to the PSII workflow tutorial.

PR #751, opened 2021-04-13
updated the PSII tutorial with segmentation recommendations

annacasto:
I've updated the PSII tutorial documentation to recommend using the Fmin image to choose a threshold for segmentation because Fmax images have some noise around the leaf borders, possibly due to leaf movement. This problem is described more in issue #749. I also added a sentence to recommend using the erode function after the 'fill' step if it's not possibly to chose a threshold using Fmin. I updated corresponding figures as well.


HaleySchuhl (approved):

docs/psII_tutorial.md
```diff
@@ -110,15 +110,16 @@ def main():
...
 
-**Figure 4.** Rotated `fmax` frame  
+**Figure 4.** Rotated `fmin` frame  
 
-![Screenshot](img/tutorial_images/psII/fmax_rescaled.jpg)
+![Screenshot](img/tutorial_images/psII/fmin2.png)
 
 The resulting image is then thresholded with a [binary threshold](binary_threshold.md) to capture the plant material. In most cases, it is expected that pixel values
 range between 0 and 255, but our example image has pixel values from 0 to over 7000. Trial and error is a common method for selecting an appropriate threshold value.
+We recommend choosing a threshold based on the Fmin image to reduce noise that may be introduced by the bright imaging light used when capturing Fmax. If this is not possible, we recommend using the [erode](erode.md) function after the [fill](fill.md) step to reduce noise around the plant border.
```
dschneiderch, line 119: I think it would be better to say something like:
... and in different imaging scenarios it may be easier to segment Fmin or Fmax. Fmax will have higher dynamic range to more easily pick a precise threshold value but Fmin may have less errant signal, e.g. if the leaf was moving during imaging.

annacasto: Yeah, I like that wording better. I think it makes the options clearer. I'll update the text.

nfahlgren:
Thanks @annacasto!

PR #757, opened 2021-04-20
Update morphology fill segments

DannieSheng:
**Describe your changes**
Add a `colorize_label_img` function to the `visualize` sub-package that colorize input label image. 

Update the fill_segments function in the morphology sub-package such that the added observations are corrected. 
Also return the `filled_mask` (which is a label image as an output) along with the `filled_image` as outputs,
Also, the filled_img is generate by calling the added `colorize_label_img` function. 

**Type of update**
Is this a:
* Bug fix
* Feature enhancement
* Update to documentation

**Associated issues**
#755


HaleySchuhl (approved):

mkdocs.yml
```diff
@@ -173,6 +173,7 @@ nav:
           - 'Histogram': visualize_histogram.md
           - 'Overlay Images': visualize_ovelay_two_imgs.md
           - 'Pseudocolor': visualize_pseudocolor.md
+          - 'Colorize Label Image': visualize_colorize_label_img.md
```
nfahlgren, line 176: @DannieSheng we have generally been listing the functions alphabetically by the title in the table of contents (with some exceptions). Could you move this one up above 'Colorize Masks'?

DannieSheng: Yes of course! Thank you!

plantcv/plantcv/morphology/fill_segments.py
```diff
@@ -55,21 +55,24 @@ def fill_segments(mask, objects, stem_objects=None, label="default"):
...
+
+    debug = params.debug
+    params.debug = None
+    filled_img = colorize_label_img(filled_mask)
+    params.debug = debug
     _debug(visual=filled_img, filename=os.path.join(params.debug_outdir, str(params.device) + "_filled_img.png"))
 
-
-    return filled_img
+    return filled_img, filled_mask
```
nfahlgren, line 78: Even though it's sort of a breaking change, I was thinking that rather than returning `filled_img` we would return `filled_mask` instead, only. But we would display `filled_img` in debug mode. And if the user wants to create the `filled_img` they can use the new `colorize_label_img` function. What's your opinion @DannieSheng and @HaleySchuhl?

DannieSheng: Did you mean that we generate the image and show `filled_img` in `plot` debug mode, but just not return it?

nfahlgren: Yeah, I'm proposing the code stays as you have it but that we only `return filled_mask`

DannieSheng: Well that makes sense to me. Users can view / save debug `filled_img` in `plot` / `print` mode, the `filled_img` itself might not be super useful in other calculations.

PR #777, opened 2021-06-08
Add a module "ecdf" (Empirical Cumulative Distribution Function) to "visualize" sub-package

DannieSheng:
**Describe your changes**
Add a new way to visualize: empirical cumulative distribution function (eCDF)
- A new function called "obj_size_ecdf" is added to the "visualize" sub-package

**Type of update**
Is this a:
* New feature
* Update to documentation
* Work in progress

Associated issues
#778

PR #778, opened 2021-06-08
Add visualize sizes

HaleySchuhl:
**Describe your changes**
Adds a function for annotating the sizes of separate objects onto a visualization. Input image should be the same size as the binary mask. Also includes a `num_objects` parameter (default set to 100 for now) that limits the number of objects that get their sizes annotated (to avoid crowded, unreadable visualizations). 

Also updated the debug method (and made respective changes to the tests) for `pcv.visualize.colorize_masks` and `pcv.visualize.auto_threshold_methods`. 

**Type of update**
Is this a:

**Associated issues**
#777 

**Additional context**


HaleySchuhl:
I got the functionality to include a number limit, where the X largest contours will get annotated. Right now, those are also the only contours that will get colored. Does this seem appropriate? I could instead color all contours found and just annotate the X largest. 

image

nfahlgren:
I think the way you have it makes sense. An upside is that you don't run out of differentiable colors as fast

docs/visualize_sizes.md
```diff
@@ -0,0 +1,44 @@
...
+
+pcv.params.debug='plot'
+
+plotting_img = pcv.visualize.sizes(img=rgb_img, mask=mask, num_objects=27)
+
+```
+
+**Output Visualization**
+
+![Screenshot](img/documentation_images//visualize_sizes/annotated_leaf_sizes.jpg)
```
DannieSheng, line 41:
```suggestion
![Screenshot](img/documentation_images/visualize_sizes/annotated_leaf_sizes.jpg)
```

DannieSheng (commented):
I think everything looks great. I do think there are two points we might want to think of:

1. The color of the text:
I happened to test it with an image with a gray background. And it seems that the default color won't display well on this image
image

2. The size of the text. With default settings of `pcv.params.text_size`(0.55), I was unable to see the texts on the image. I think mentioning it in the documentation would be nice.

plantcv/plantcv/visualize/sizes.py
```diff
@@ -0,0 +1,76 @@
...
+import cv2
+import random
+import numpy as np
+# from plantcv.plantcv import find_objects
+from plantcv.plantcv import params
+from plantcv.plantcv import color_palette
+from plantcv.plantcv._debug import _debug
+
+
+def sizes(img, mask, num_objects=100):
```
nfahlgren, line 13: I'm thinking `pcv.visualize.sizes` might not quite be descriptive enough? What about `pcv.visualize.obj_sizes` maybe?

DannieSheng: `pcv.visualize.obj_sizes` looks good to me

HaleySchuhl: Agreed!

nfahlgren (approved):
Thanks @HaleySchuhl!

PR #782, opened 2021-06-15
Add codacy config file

nfahlgren:
**Describe your changes**
Adds a configuration file for [Codacy](codacy.com). Allows fine-tuning of Python code quality checks.

**Type of update**
Is this a: Work in progress

### DIFF
--- a/.codacy.yml
+++ b/.codacy.yml
@@ -1,0 +1,9 @@
+---
+engines:
+  prospector:
+    pep8:
+      options:
+        max-line-length: 120
+    pep257:
+      disable:
+        - D202


dschneiderch:
I like the control this gives. It'd be great if we could invert PEP257 D202 :)  Line  after docstring required!
Also, It appears that for every function that has a description you are supposed to start on the 2nd line (D213).

I'd propose working towards this model:
```
def function(x):
"""
Summary

Description
"""

code starts here
```

nfahlgren:
We could merge this into the main branch and see how it goes on the backend for awhile. It shouldn't hurt anything at the moment since we don't have codacy as a blocking requirement

nfahlgren:
Maybe to your point on Tuesday, we should not turn off D202 as I thought?

dschneiderch:
up to you. I'd turn it off so I could place a newline after the docstring but it shouldn't be an issue to leave it on and remove the blank line between the second `"""` and the first line of code

nfahlgren:
I turned off the Markdown linter. I could have looked into it more but it seemed like one of those things where without configuration you couldn't satisfy it. You always have either too few or too many spaces in list items (for example).

PR #784, opened 2021-06-16
Use colour-science package for wavelength conversion

nfahlgren:
**Describe your changes**
@DannieSheng this PR is an updated version of the visualize_hyper_histogram branch that uses a different method for converting wavelengths to RGB color values. I'm a little concerned about using the code from http://www.noah.org/wiki/Wavelength_to_RGB_in_Python because the license is not clear.

As an alternative, I found the package `colour-science` (https://www.colour-science.org/) that has a lot of cool features and can convert wavelengths to other color models. The code here uses it instead for you to review. It can convert wavelengths between 390 and 830 nm. I'm not sure it's still working with the non-visible wavelengths like you had it before:

Screen Shot 2021-06-15 at 11 31 19 PM

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -15,6 +15,7 @@ dependencies:
   - dask-jobqueue
   - nb_conda
   - opencv<4, >=3.4
+  - colour-science
 channels:
   - conda-forge
   - defaults
--- a/plantcv/plantcv/visualize/hyper_histogram.py
+++ b/plantcv/plantcv/visualize/hyper_histogram.py
@@ -5,87 +5,11 @@ import numpy as np
 import pandas as pd
 from plantcv.plantcv._debug import _debug
 from plantcv.plantcv import fatal_error, params, color_palette
-from plotnine import ggplot, aes, geom_line, scale_color_manual
+from plotnine import ggplot, aes, geom_line, scale_color_manual, theme_classic
 from plantcv.plantcv.visualize import histogram
 from plantcv.plantcv.hyperspectral import _find_closest
 from scipy.spatial import distance
-
-'''
-    == A few notes about color ==
-
-    Color   Wavelength(nm) Frequency(THz)
-    Red     620-750        484-400
-    Orange  590-620        508-484
-    Yellow  570-590        526-508
-    Green   495-570        606-526
-    Blue    450-495        668-606
-    Violet  380-450        789-668
-
-    f is frequency (cycles per second)
-    l (lambda) is wavelength (meters per cycle)
-    e is energy (Joules)
-    h (Plank's constant) = 6.6260695729 x 10^-34 Joule*seconds
-                         = 6.6260695729 x 10^-34 m^2*kg/seconds
-    c = 299792458 meters per second
-    f = c/l
-    l = c/f
-    e = h*f
-    e = c*h/l
-
-    List of peak frequency responses for each type of 
-    photoreceptor cell in the human eye:
-        S cone: 437 nm
-        M cone: 533 nm
-        L cone: 564 nm
-        rod:    550 nm in bright daylight, 498 nm when dark adapted. 
-                Rods adapt to low light conditions by becoming more sensitive.
-                Peak frequency response shifts to 498 nm.
-'''
-
-
-def _wavelength_to_rgb(wavelength, gamma=0.8):
-    """
-    reference: "Wavelength to RGB in Python." Noah.org, . 20 Sep 2014, 06:38 UTC. 1 Apr 2021, 19:06
-    <https://www.noah.org/mediawiki-1.34.2/index.php?title=Wavelength_to_RGB_in_Python&oldid=7665>.
-    http://www.noah.org/wiki/Wavelength_to_RGB_in_Python
-
-    This converts a given wavelength of light to an approximate RGB color value. The wavelength must be given in
-    nanometers in the range from 380 nm through 750 nm (789 THz through 400 THz).
-
-    Based on code by Dan Bruton
-    http://www.physics.sfasu.edu/astro/color/spectra.html
-
-    Inputs:
-    wavelength: wavelength of a visible wavelength in nanometers range: (380, 750)
-    gamma: (optional) gamma correction. default value = 0.8
-
-    Returns
-    :param wavelength: float
-    :param gamma: (optional) float
-    :return: tuple
-    """
-
-    wavelength = float(wavelength)
-    if 380 <= wavelength <= 440:
-        attenuation = 0.3 + 0.7 * (wavelength - 380) / (440 - 380)
-        r, g, b = ((-(wavelength - 440) / (440 - 380)) * attenuation) ** gamma, 0.0, (1.0 * attenuation) ** gamma
-    elif 440 <= wavelength <= 490:
-        r, g, b = 0.0, ((wavelength - 440) / (490 - 440)) ** gamma, 1.0
-    elif 490 <= wavelength <= 510:
-        r, g, b = 0.0, 1.0, (-(wavelength - 510) / (510 - 490)) ** gamma
-    elif 510 <= wavelength <= 580:
-        r, g, b = ((wavelength - 510) / (580 - 510)) ** gamma, 1.0, 0.0
-    elif 580 <= wavelength <= 645:
-        r, g, b = 1.0, (-(wavelength - 645) / (645 - 580)) ** gamma, 0.0
-    elif 645 <= wavelength <= 750:
-        attenuation = 0.3 + 0.7 * (750 - wavelength) / (750 - 645)
-        r, g, b = (1.0 * attenuation) ** gamma, 0.0, 0.0
-    # else:
-    #     R, G, B = 0.0, 0.0, 0.0
-    r *= 255
-    g *= 255
-    b *= 255
-    return int(r), int(g), int(b)
+import colour
 
 
 def _rgb_to_webcode(rgb_values):
@@ -146,9 +70,9 @@ def hyper_histogram(array, mask=None, bins=100, lower_bound=None, upper_bound=No
     match_ids = [_find_closest(wls, wv) for wv in wvlengths]
 
     # check if in the visible wavelengths range
-    ids_vis = [idx for (idx, wv) in enumerate(wvlengths) if 380 <= wv <= 750]
+    ids_vis = [idx for (idx, wv) in enumerate(wvlengths) if 390 <= wv <= 830]
     # invisible wavelengths
-    ids_inv = [idx for (idx, wv) in enumerate(wvlengths) if wv < 380 or wv > 750]
+    ids_inv = [idx for (idx, wv) in enumerate(wvlengths) if wv < 390 or wv > 830]
 
     colors = [tuple(x) for x in color_palette(len(wvlengths))]
     if len(ids_vis) < len(wvlengths):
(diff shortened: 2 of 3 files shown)


DannieSheng:
> **Describe your changes**
> @DannieSheng this PR is an updated version of the visualize_hyper_histogram branch that uses a different method for converting wavelengths to RGB color values. I'm a little concerned about using the code from http://www.noah.org/wiki/Wavelength_to_RGB_in_Python because the license is not clear.
> 
> As an alternative, I found the package `colour-science` (https://www.colour-science.org/) that has a lot of cool features and can convert wavelengths to other color models. The code here uses it instead for you to review. It can convert wavelengths between 390 and 830 nm. I'm not sure it's still working with the non-visible wavelengths like you had it before:
> 
> Screen Shot 2021-06-15 at 11 31 19 PM

I think it looks good to me. The way I convert from wavelengths to rgb tuples doesn't work for non-visible wavelengths either. 
I'll merge it and test it.

DannieSheng (approved):

DannieSheng:
I was testing the functionality of the new color-coding method. However, I only found this issue, which may make this tool not ideal. 
image
As shown in the plot, by default, the histograms of red, green, and blue are shown. However, using the color-coding method provided by `colour-science` makes the red color appears darker than we expected. I did some research and potentially we can make the default wavelength for red color to be 650nm (instead of 670 nm).
To obtain a tuple of RGB values from a given wavelength value, normally 2 steps are involved:
1. Calculate the XYZ values from wavelength: wavelength -> XYZ
2. Calculate the RGB values from XYZ values: XYZ to RGB
I looked at also [this](https://scipython.com/blog/converting-a-spectrum-to-a-colour/) reference and found that there is no big difference in step 1. However, I suspect that the differences in step 2 cause the difference in final values. To be specific, there is no normalization in the `colour-science` before returning the RGB values. Shown in the example below:
wavelength = 670nm
- rgb value using `colour-science`: (0.21898368, -0.01999074, -0.00193949).
But we set negative values to zero before scaling          
```
# Set negative values to zero before scaling
rgb[np.where(rgb < 0)] = 0
```
In other words, the rgb value becomes (0.21898368,  0. ,0.)

- rgb value using the function in the other reference: (1.,  0.  ,  0.08890881)
After converting to `uint8` datatype, the RGB values are (56,0,0) and (255,0,23), respectively.

image

image

The former one is also red in color but much darker.

@nfahlgren what do you think

nfahlgren:
Based on the link you sent it sounds like negative RGB values occur when the calculated value is outside of the sRGB region. It sounds like this can be corrected by adjusting the chromaticity coordinates, which I think are the illuminant settings in the `wavelength_to_XYZ` function. It doesn't make 670nm brighter but I think I had nonsensical values and that these are more appropriate:

```python
illuminant_XYZ=np.array([0.34570, 0.35850])
 illuminant_RGB=np.array([0.31270, 0.32900])
```

Then also on the link you sent, I think their method for adjusting for negative RGB values makes more sense than setting all negative values to zero, so I tested this update to our code:

```python
# rgb[np.where(rgb < 0)] = 0
if np.any(rgb < 0):
    cor = -np.min(rgb)
    rgb += cor
if not np.all(rgb==0):
    # Normalize the rgb vector
    rgb /= np.max(rgb)
```

The normalization step at the end brightens the colors a lot but makes the colors look too similar. Maybe we could do a gamma correction instead?

DannieSheng:
I tested using gamma correction, I am still not very satisfied with the result:
image
Still using the example with wavelength=670nm. The RGB tuple is (0.22705707,  0.,  0.01809794), after gamma correction with `gamma=0.8`, the RGB tuple becomes (0.30542742,  0.,  0.04037414). After transforming the datatype into `uint8`, the tuple becomes (58,0,5). And I think this causes none of the bands to be saturated, hence the "darkness" of the final color.

nfahlgren:
How about this?

```python
# Help visualize histograms for hyperspectral images

import os
import sys
import numpy as np
import pandas as pd
from plantcv.plantcv._debug import _debug
from plantcv.plantcv import fatal_error, params, color_palette
from plotnine import ggplot, aes, geom_line, scale_color_manual, theme_classic
from plantcv.plantcv.visualize import histogram
from plantcv.plantcv.hyperspectral import _find_closest
from scipy.spatial import distance
import colour


def _wavelength_to_rgb(wvlength, cmfs = colour.MSDS_CMFS["CIE 2012 10 Degree Standard Observer"],
                       matrix = np.array([[3.24062548, -1.53720797, -0.49862860],
                                          [-0.96893071, 1.87575606, 0.04151752],
                                          [0.05571012, -0.20402105, 1.05699594]])):
    """
    Converts wavelength to RGB tuple, using the "colour" package
    :param wvlength:
    :param cmfs:
    :param matrix:
    :return:
    """


    # convert from wavelength to xyz
    xyz = colour.wavelength_to_XYZ(wvlength, cmfs)
    illuminant_xyz = np.array([0.34570, 0.35850])
    illuminant_rgb = np.array([0.31270, 0.32900])

    # convert from xyz to rgb
    rgb = colour.XYZ_to_RGB(xyz, illuminant_XYZ=illuminant_xyz, illuminant_RGB=illuminant_rgb,
                            chromatic_adaptation_transform="Bradford",
                            matrix_XYZ_to_RGB=matrix)
    color_vis = colour.notation.RGB_to_HEX(rgb)
    
    return color_vis


def hyper_histogram(array, mask=None, bins=100, lower_bound=None, upper_bound=None,
                    title=None, wvlengths=[480, 550, 650]):
    """This function calculates the histogram of selected wavelengths hyperspectral images
    The color of the histograms are based on the wavelength if the wavelength is in the range of visible spectrum;
    otherwise, random colors are assigned

    Inputs:
    array        = Hyperspectral data instance
    mask         = binary mask, if provided, calculate histogram from masked area only (default=None)
    bins         = divide the data into n evenly spaced bins (default=100)
    lower_bound  = the lower bound of the bins (x-axis min value) (default=None)
    upper_bound  = the upper bound of the bins (x-axis max value) (default=None)
    title        = a custom title for the plot (default=None)
    wvlengths    = (optional) list of wavelengths to show histograms (default = [480,550,670],  i.e. only show
                                                                    histograms for blue, green, and red bands)

    Returns:
    fig_hist      = histogram figure

    :param array: plantcv.plantcv.classes.Spectral_data
    :param mask: numpy.ndarray
    :param bins: int
    :param lower_bound: None, int, float
    :param upper_bound: None, int, float
    :param title: None, str
    :param wvlengths: list
    :return fig_hist: plotnine.ggplot.ggplot
    """

    # Always sort desired wavelengths
    wvlengths.sort()

    # Available wavelengths of the spectral data
    wl_keys = array.wavelength_dict.keys()
    wls = np.array([float(i) for i in wl_keys])

    # Spectral resolution of the spectral data
    diffs = [wls[i] - wls[i - 1] for i in range(1, len(wls))]
    spc_res = sum(diffs) / len(diffs)

    # Check if the distance is greater than 2x the spectral resolution
    # If the distance > 2x resolution, it is considered being out of available ranges
    checks = []
    for i in range(0, len(wvlengths)):
        checks.append(array.min_wavelength - wvlengths[i] > 2 * spc_res)
        checks.append(wvlengths[i] - array.max_wavelength > 2 * spc_res)
    if np.any(checks):
        fatal_error(f"At least one band is too far from the available wavelength range: "
                    f"({array.min_wavelength},{array.max_wavelength})!")

    # Find indices of bands whose wavelengths are closest to desired ones
    match_ids = [_find_closest(wls, wv) for wv in wvlengths]

    # Check if in the visible wavelengths range
    ids_vis = [idx for (idx, wv) in enumerate(wvlengths) if 390 <= wv <= 830]
    ids_uv = [idx for (idx, wv) in enumerate(wvlengths) if wv < 390]
    ids_ir = [idx for (idx, wv) in enumerate(wvlengths) if wv > 830]
    
    # Colors
    color_scale_tmp = params.color_scale
    # UV colors
    params.color_scale = "twilight_shifted"
    rgb_uv = color_palette(num=len(ids_uv), saved=False)
    colors_uv = []
    for rgb in rgb_uv:
        rgb = np.array(rgb).astype(float) / 255
        colors_uv.append(colour.notation.RGB_to_HEX(rgb))
    # IR colors
    params.color_scale = "inferno"
    rgb_ir = color_palette(num=len(ids_ir), saved=False)
    colors_ir = []
    for rgb in rgb_ir:
        rgb = np.array(rgb).astype(float) / 255
        colors_ir.append(colour.notation.RGB_to_HEX(rgb))
    params.color_scale = color_scale_tmp
    # VIS colors
    colors_vis = []
    for i in ids_vis:
        colors_vis.append(_wavelength_to_rgb(wvlength=wvlengths[i]))

    array_data = array.array_data

    # List of wavelengths recorded created from parsing the header file will be string, make list of floats
    histograms = dict()
    hist_dataset = pd.DataFrame(columns=['reflectance'])
    debug = params.debug
    params.debug = None
    colors = colors_uv + colors_vis + colors_ir

    # Create a dataframe for all histogram related information (using the "histogram" function in "visualization" subpackage)
    for i_wv, (wv, color) in enumerate(zip(wvlengths, colors)):
        idx = match_ids[i_wv]
        _, hist_data = histogram(array_data[:, :, idx], mask=mask, bins=bins, lower_bound=lower_bound,
                                 upper_bound=upper_bound, title=title, hist_data=True)
        histograms[wv] = {"label": wv, "graph_color": color, "reflectance": hist_data['pixel intensity'].tolist(),
                          "hist": hist_data['proportion of pixels (%)'].tolist()}
        if i_wv == 0:
            hist_dataset['reflectance'] = hist_data['pixel intensity'].tolist()
        hist_dataset[wv] = hist_data['proportion of pixels (%)'].tolist()

    # Make the histogram figure using plotnine
    df_hist = pd.melt(hist_dataset, id_vars=['reflectance'], value_vars=wvlengths,
                      var_name='Wavelength (' + array.wavelength_units + ')', value_name='proportion of pixels (%)')

    fig_hist = (ggplot(df_hist, aes(x='reflectance', y='proportion of pixels (%)',
                                    color='Wavelength (' + array.wavelength_units + ')'))
                + geom_line()
                + scale_color_manual(colors, expand=(0, 0))
                + theme_classic()
                )
    params.debug = debug
    _debug(fig_hist, filename=os.path.join(params.debug_outdir, str(params.device) + '_histogram.png'))

    return fig_hist
```

DannieSheng:
I tested it (apparently it took longer than I expected) but the displayed color is still not satisfying. I went through a bunch of information regarding calorimetry, and my current observations and understandings are:
1. If there are smaller than 0 or greater than 1 RGB values, there will be warnings when trying to convert to HEX. Those values are not wrong but are just out of the gamut of the target color space (sRGB). As shown in the figure
image
For "out of gamut" values, the `colours` would `clip` (for negative values) or `max_normalize` (for greater than 1 values)
Letting the `colours` normalize or clip values would cause the RGB values to be "too small" (not saturated). e.g. I got a lot of tuples like (2.48399407, -0.20860912, -0.01999058)

2. [This discussion](https://github.com/colour-science/colour/discussions/746) talks about the same thing. And I think the suggestion is that using the building normalization (`clip` and `max_normalize`). Also in [this discussion](https://github.com/colour-science/colour/discussions/711) it is pointed out that the terminology is confusing: "Input *RGB* colourspace array is in domain [0, 1] suggests "constrained to the range 0-1" but really it means it is normalized to that range."

3. Another issue not related to saturation is that, with a spectral resolution of 1nm, I got the same HEX color code for a number of wavelengths. It would be an issue because the spectral resolution for our hyperspectral images is about 0.6. Although it is unlikely if someone tries to display the histogram for neighboring bands.

nfahlgren:
Right, but what's the issue that makes this not satisfying?

DannieSheng:
The plotted lines are dark, similar to the previous result above

nfahlgren:
Screen Shot 2021-06-23 at 1 23 05 PM

These don't seem dark to me?

nfahlgren:
There will always be an issue plotting distinct colors when a large number of wavelengths are plotted I think

nfahlgren:
Screen Shot 2021-06-23 at 1 28 01 PM

This is another way to look at the colors the package can plot from spectrum

DannieSheng:
> Screen Shot 2021-06-23 at 1 23 05 PM
> 
> These don't seem dark to me?

Did the code above help you generate this plot?

nfahlgren:
I used the `hyper_histogram` function above to make the histogram plot

DannieSheng:
I see the problem. I did not get a result similar to yours, that's the reason I was not satisfied.

nfahlgren:
Oh, is it behaving differently on your system versus mine?

DannieSheng:
It seems to be. I am suspicious that I made some mistakes when copying over the code. However, I am also wondering which test image you used. It seems that all the reflectances are really small.

PR #790, opened 2021-06-18
update photosynthesis package to use xarray and analyze entire induction curve protocols

### LINKED ISSUE #636: Photosynthesis analysis quality filtering
When we calculate a metric like `(Fm - F0) / Fm`, there can be some quality control issues. One that came up is that when `F0 = 0` then `Fv/Fm = 1`. I think this is not a productive result both because it lacks biological relevance and because our lack of ability to detect signal in F0 could just be a detection limit issue.

I propose that we only calculate Fv/Fm where F0 > 0 within the `plantcv.photosynthesis.analyze_fvfm` function.

Alternatively, we do not change anything and leave it to the user to filter the results. I worry a little about creating a non-transparent process (though we would document it) but I think this is a non-controversial filter to add?

cc: @annacasto, @dschneiderch 

**Additional context**
If we do add quality filtering, this could also apply to other photosynthesis functions being developed in #426.

dschneiderch:
This pull request implements the new data model for psII measurements and associated analyze_* functions.
We are using a new class to store multiple data arrays. Each data array has 4 dimensions (x, y, frame_label, measurement). Each data array must have a predefined name: darkadapted, lightadapted, rgb, chlorophyll, anthocyanin

[x] PSII_data class to store multiple xarray data arrays
[x] read_cropreporter to import data from cropreporter instrument
[] read_walz to import data from walz imaging pam
[x] analyze_yii to analyze photosynthetic efficiency for any number of measurements 
[x] analyze_npq to analyze nonphotochemical quenching for any number of measurements
[x] tests
[x] documentation

**Type of update**
Is this a:
* Work in progress

**Associated issues**
closes #636 

**Additional context**


dschneiderch:
Woo! I think its finally ready for a review with 2 caveats:
1. I don't think xarray.plot() respects pcv.params.debug=None and I'm not sure how to deal with that
2. the histograms from analyze_yii() and analyze_npq() need to be combined - if there are multiple measurements then only the last histogram is output.  xarray accepts matplotlib subplots arguments but I haven't managed to wrap my brain around it.

nfahlgren:
Awesome, thanks @dschneiderch!

plantcv/plantcv/photosynthesis/read_cropreporter.py
```diff
@@ -2,83 +2,172 @@
...
+            )
+            rgb.attrs["long_name"] = "rgb measurements"
+            ps.add_data(rgb)
+
+            params.debug = debug
+            _debug(visual=ps.rgb,  # does it make sense to show all three bands separately?
+                   filename=os.path.join(params.debug_outdir, f"{str(params.device)}_CLR-RGB.png"))
+
+        elif corresponding_dict[key] == "CHL":
+            frame_labels = ["Chl", "Chl-NIR"]
```
HaleySchuhl, line 146: switch, NIR first frame ?

plantcv/plantcv/photosynthesis/read_cropreporter.py
```diff
@@ -2,83 +2,172 @@
...
+                name='chlorophyll'
+            )
+            chl.attrs["long_name"] = "chlorophyll measurements"
+            ps.add_data(chl)
+
+            _debug(visual=ps.chlorophyll,
+                   filename=os.path.join(params.debug_outdir, f"{str(params.device)}_CHL-NIR.png"))
+
+        elif corresponding_dict[key] == "SPC":
+            frame_labels = ["Anth", "Far-red", "Anth-NIR"]
```
HaleySchuhl, line 160: swap orderl NIR, far-red, then Anth

plantcv/plantcv/classes.py
```diff
@@ -217,3 +216,28 @@ def __init__(self, array_data, max_wavelength, min_wavelength, max_value, min_va
...
+
+class PSII_data:
+    """PSII data class"""
+
+    def __init__(self):
+        self.darkadapted = None
+        self.lightadapted = None
+        self.rgb = None
+        self.anthocyanin = None
+        self.chlorophyll = None
```
nfahlgren, line 229: @dschneiderch and @HaleySchuhl any objections if I add class attributes for `datapath` and `filename` (or similar). Then I would update `read_cropreporter` to return only `ps`. This would be consistent with where we want to move to with other input data

nfahlgren:
The debug images are histograms but I'm not sure what they show?

plantcv/plantcv/classes.py
dschneiderch: No, I think thats better.
Wouldn't it be an instance attribute though?

nfahlgren: Cool, yep. I think I would just define them in the class like the others with a default value of `None` and then assign values in the instance.

dschneiderch:
> The debug images are histograms but I'm not sure what they show?

which file/function?

nfahlgren:
Oh right sorry, the debug output for read_cropreporter

nfahlgren:
I just pushed a big refactoring for `read_cropreporter`. The reason for the new structure is mainly that in discussion with one of our students, we decided that combining the CLR (RGB) and SPC (Green, Far-red, NIR) datasets into a `Spectral_data` object would work best with how we need to analyze that data downstream (calculating anthocyanin reflectance index and chlorophyll index using functions that already exist in `pcv.spectral_index`). Since I need to combine up to two DAT files for that I ended up making private functions to handle each dataset rather than looping over them. 

The only awkward thing (not related to the structure of the function itself) is that the PSII_data object will be a mix of mostly all xarray DataArrays and one Spectral_data object. But probably okay. In the future we might change the way we handle hyperspectral data and use xarray instead and then it would all be the same.

Note: coverage will probably drop a little because I didn't update the tests yet.

HaleySchuhl (approved):

dschneiderch:
@nfahlgren i pushed an update to read_cropreporter to show the frames as images in debug. I think thats what I meant to do but forgot about it after several refactoring attempts. however, debug=None still does not suppress the plots

nfahlgren:
If we move DataArray.plot methods to `plot_image` and `print_image` by using `kwargs` in both then it mostly obeys `pcv.params.debug`. The only caveat is that in print mode it will still display the plots if there is a plotting device

nfahlgren:
In `analyze_yii` the pcolormesh plot throws an error (even with the old plot_image code). @dschneiderch do you understand why this works with the `read_cropreporter` DataArrays:

```python
da.plot(col='frame_label', col_wrap=4)
```

But this doesn't work with `analyze_yii`?

```python
yii.plot(ol='measurement', col_wrap=4)
```

Because if that did it would all play together nicely

dschneiderch:
Is the typo real?

Arg should be col, not ol
________________________________

nfahlgren:
Just a copy and paste error here, I had updated the comment but probably didn't push to the email notification

dschneiderch:
honestly not sure. I think it is complaining because measurement only has length 1 and it is getting squeezed to length 0. If you use a dataset with multiple measurement values it works. Personally I think this use case makes sense programmatically work so maybe its a bug. I will bring it up upstream.

A couple options for now. 
1.
we could make `plot_image()` always display an image, forgoing any of their [other options](http://xarray.pydata.org/en/stable/generated/xarray.DataArray.plot.html)
`img.plot.pcolormesh(**kwargs)`
and then using 
```
_debug(img=ps.yii,
filename=....,
col='measurement',
col_wrap=int(np.ceil(img.measurement.size)/4), 
robust=True)
```

2. add some logic in plot_image to see how many measurement values there are and then plot accordingly.


Oddly, this works for 1 measurement but not more:
```
fvfm.plot(col_wrap='measurement')
```
[`col_wrap` is supposed to be an integer though...](http://xarray.pydata.org/en/stable/generated/xarray.DataArray.plot.html)  and `col` is supposed to be the dim, but that doesn't work as you found out. presumably this will stop working at some point.

dschneiderch:
here's the issue [Don't squeeze DataArray before plotting](https://github.com/pydata/xarray/issues/620)

nfahlgren:
Using `FacetGrid` directly seems to work for both one and multiple measurements:

```python
gd = FacetGrid(data=fvfm, col="measurement", col_wrap=1)
gd.map_dataarray(xr.plot.imshow, "x", "y")
```

We could put this directly in the YII and NPQ functions for now and not use the `_debug` module.

dschneiderch:
@nfahlgren next level wizardry!
another idea I had overnight is to check whether dims x and y are present and if so call pcolormesh(). i actually think they do something like this under the hood but then squeeze the dims too early. I think this would cover most use cases.

dschneiderch:
also, for hte col_wrap arg I suggest `col_wrap=int(np.ceil(fvfm.measurement.size/3))` where the number is the max # of columns

dschneiderch:
this will all fail if the `col` arg is removed from this test because then _show_dataarray() returns a histogram as a tuple which does not have `self.fig.savefig()`

should we just always limit to pcolormesh()?   what does a line plot return from xarray?

```
def test_plantcv_print_image_dataarray():
    da = psii_cropreporter('darkadapted').squeeze('measurement', drop=True)
    try:
        #this fails without col or row as an arg because xarray will create a matplotlib histogram that comes back as a tuple and then print_image will try to use self.fig.savefig which doesn't exist for a tuple
        pcv.print_image(img=da, col='frame_label', filename='/dev/null') 
    except RuntimeError:
        assert False
    # Assert that the image was plotted without error
    assert True
```

dschneiderch:
@nfahlgren   I think I may have broken the debug= functionality again. sometimes I see the plots even when I have debug=None or debug='print'. clearly I don't understand something.

however, in terms of plotting data array data, I think it will be best to limit to facetgrid functionality using pcolormesh(). it greatly simplifies print_image and plot_image from what I can tell.
the plot output from xarray is different based on the number of dimensions:
1D: da.plot() as a list len 1 containing matplotlib.lines.Line2D bc xarray makes a line plot
3+D: da.plot() comes back as a tuple bc xarray makes a histogram unless `col` and/or `row` is supplied for facetting
2D: da.plot() will make an image. facetted images if extra dims are specified as kwargs `col` or `row`. in practice I found it's easy to lose track of this and provide too many or not enough dims so I explicitly call pcolormesh() with some sanity checks/error messages in a new private module `_show_dataarray` . I also check for x and y because I think thats the main use case

nfahlgren:
I agree, limiting to pcolormesh makes sense for our usecase

nfahlgren (approved):
Alright, I think everything is wrapped up! Thanks for the huge effort on this @dschneiderch and @HaleySchuhl!

PR #791, opened 2021-06-18
Match multiple metadata values - simpler approach

bganglia:
**Describe your changes**

This is a replacement to pull request #520.

It supports one syntax at the command line (``--match id:1,id:2,id:3``), and also supports lists in config.json.

Any error messages PlantCV currently gives for incorrect input should still appear, since the code for reading metadata filters is mostly the same.

I wasn't sure where to put tests for ``plantcv-workflow.py`` itself. At the moment these are in ``tests/tests.py``, but let me know if you use a different way for running these tests.

**Type of update**

**Associated issues**
This pull request addresses issue #461.

**Additional context**
This PR is a replacement to #520.


dschneiderch:
hi @bganglia, thanks for the followup!  the changes in the code look good and everything works for me. all tests are in tests.py but i'd need to sort through it to figure out how to test these (complex!) functions.  maybe @nfahlgren can point you to the right place.

bganglia:
Thanks.

The tests use the images like ``NIR_TV_0_z300_h1_g0_e15000_v500_93059.png`` in ``tests/data`` with these cases:

``imgtype:VIS,camera:SV``
``imgtype:VIS,imgtype:NIR,camera:TV``
``imgtype:VIS,imgtype:NIR,camera:SV,camera:TV``

There is a different test for using a config file.

https://github.com/danforthcenter/plantcv/blob/758c8d4e6b35cd4b6c69d14a79aa85d395fa1b61/tests/tests.py#L883-L969

I used subprocess to call plantcv-workflow.py, but if there is a more plantcv-style way to do this, that would also work.

nfahlgren:
Because of code duplication in my original `metadata_parser` function, I had to duplicate the multi-value checking code introduced here in two other places to work for both types of dataset structures we use. But that complicated testing so I partially fixed my original mess by making a private function out of the metadata filter checking code. 

Looks like it's passing now, thanks for the update!

nfahlgren (approved):

PR #797, opened 2021-07-09
Add Anth and Chl anaylsis functions

jwheeler5:
**Describe your changes**
Adds functions for analyzing CropReporter anthocyanin and chlorophyll data

**Type of update**
* Update to documentation
* Work in progress

**Associated issues**

**Additional context**


HaleySchuhl:
- [ ] update `read_cropreporter` function
     - change the order of color channels to blue, green, red so that they are in spectral order. This will later make it possible to integrate hyperspectral abilities. 
     - don't convert to 8-bit (remove 128-130) for color data, instead make reading this part in the exact same as SPC (`data=img_cube`) 
- [ ] analysis functions should take the entire PSII instance and then pull out the frames needed.

dschneiderch:
> * [ ]  analysis functions should take the entire PSII instance and then pull out the frames needed.

our discussion yesterday concluded this didn't make sense because you'd have to add a flag to `analyze_yii()` to know what to compute yii for. overall it'd be more consistent to always input dataarray's to these functions. 
But as long as you add an argument to `analyze_yii` it doesn't matter that much to me.

nfahlgren:
Closing since the functionality was alternately added in #790

PR #798, opened 2021-07-13
update watershed.py

adrianethompson:
**Describe your changes**
Changed import for watershed from skimage.morphology (which is being deprecated in an upcoming update) to skimage.segmentation, so that warnings in jupyter notebooks will go away

**Type of update**
* Bug fix

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -8,7 +8,7 @@ dependencies:
   - pandas
   - python-dateutil
   - scipy
-  - scikit-image
+  - scikit-image>=0.13
   - scikit-learn
   - plotnine
   - dask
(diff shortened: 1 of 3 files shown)


nfahlgren:
Thanks @adrianethompson!

It looks like scikit-image introduced the relocated watershed function in version 0.13. Because importing it from the new location is not backwards compatible with older versions of scikit-image I think we should add a constraint to our dependency requirements.

Could you update `requirements.txt` and `environment.yml` to update the `scikit-image` lines to `scikit-image>=0.13`?

nfahlgren (approved):
Thanks @adrianethompson!

PR #811, opened 2021-08-05
Fix line endings

nfahlgren:
**Describe your changes**
Updates gitattributes to exclude common binary files from end-of-line normalization. Normalizes line endings for existing files that were checked in with CRLF line endings and replaces them with LF line endings.

**Type of update**
Is this a: Bug fix

**Additional context**
https://docs.github.com/en/get-started/getting-started-with-git/configuring-git-to-handle-line-endings


JorgeGtz (approved):

nfahlgren:
Thanks for bringing up this issue @JorgeGtz!

PR #810, opened 2021-08-04
Pixel scatter visualization

JorgeGtz:
**Describe your changes**
Added the pixel scatter visualization that plots the pixels in one or more color images in a 2D plane formed by the specified channels. 

**Type of update**
Is this a:
* New feature or feature enhancement

**Associated issues**
Discussion #724 

**Additional context**

PR #823, opened 2021-09-15
Two io functions

JorgeGtz:
**Describe your changes**
Started the subpackage I/O with two functions that were useful for the pixel scatter visualization:
*  read_dataset - Returns a list of paths to the images in a given directory.
*  random_subset - Returns a subset of the elements in a list.

**Type of update**
Is this a:

**Associated issues**
Discussion #724 and PR #810 

**Additional context**


nfahlgren:
@JorgeGtz and @HaleySchuhl, not to overly complicate things, but what do you think about using the `pcv.io.read_dataset` method here in the timelapse video function in #762? That would just mean we would need to get this merged first and then refactor #762 somewhat.

JorgeGtz:
@nfahlgren I can do the refactoring of #762 after this it merged.

HaleySchuhl (approved):

Issue #828, opened 2021-11-01
Colors are not real after applying correction


alihamidisepehr:
**Describe the bug**
I'm using [this tutorial](https://plantcv.readthedocs.io/en/v3.11.0/transform_color_correction_tutorial/) to correct colors in photos from melons (Scenario B). Since colorchecker was tilted in some of my images, I tried to manually check  if the mask is located in color chips. I followed all the steps but images after correction were either too bright or in some cases too dark. I also used the regression plot you demonstrated at the end of the tutorial and I also got a very high correlation.

**To Reproduce*
![IMG_2153](https://user-images.githubusercontent.com/43645711/139707289-4bb10875-36df-44b2-852e-3ee6b1fef131.JPG)
![IMG_2176](https://user-images.githubusercontent.com/43645711/139707303-91bc4712-ba9a-4de6-b6a5-ca820f49456e.JPG)
![image](https://user-images.githubusercontent.com/43645711/139707928-7383d62f-f7d0-4f35-a74a-ada6104f4ce4.png)

I was expecting each color chip would match between target and source (blue to blue, red to red) after correction but I don't think that's what's happening here. Could you please help me with this?

Also, can you please explain what are the recommended criteria for a target image? Do I have to have a perfect image with only colorchecker in it? if so, can I crop one of my photos to generate a target image?

Thank you in advance!

alihamidisepehr:
Hi @nfahlgren, I was wondering if I should wait for any response here or there's going to be an update in the code? Thank you

nfahlgren:
Hi @alihamidisepehr,

Sorry, this was automatically closed when we merged the pull request with the bugfix. We have had a few people report issues with color correction with similar issues to what you reported above, thank you! @JorgeGtz identified a bug in the module and developed the fix for it in #829. He tested it with your images and it looks like it corrects the issue.

I created a bugfix release (v3.13.1) that includes the updates. It's available on PyPI already and can be installed via `pip` and should be available through `conda` in a few hours.

alihamidisepehr:
Thank you @nfahlgren. I'm looking forward to testing it with new edits. In my images I have images with different lighting conditions and in some of them there's also shadows caused by the colorspace on the lower surfaces including the side of the melon. Do you recommend using LAB colorspace or I should stay with RGB?

Thank you,
Ali

alihamidisepehr:
[...]

PR #829, opened 2021-11-03
Update color correction

JorgeGtz:
**Describe your changes**
Cast the source and target images as float64 and normalize between 0-1 (the normalization is not essential). 
This should avoid creating artifacts in the images due to integer overflow.

**Type of update**
Is this a:
* Bug fix

**Associated issues**
#828 

**Additional context**


nfahlgren (approved):
Thanks @JorgeGtz!

PR #830, opened 2021-11-04
cast contour tuple as a list in find_objects

JorgeGtz:
**Describe your changes**
Cast the contours output from cv2.findContours as a list (they now are given as a tuple) 

**Type of update**
Is this a:
* Bug fix

**Associated issues**

**Additional context**
Some morphology tests failed in the unrelated PR #829. 
The problem seems to be that in the version of OpenCV used by the tests in GitHub (3.4.16.57) the contours in findContours
are given as a tuple instead of as a list. Those tests pass locally using OpenCV 3.4.14.


nfahlgren (approved):
Thanks @JorgeGtz!

PR #850, opened 2022-01-04
Reorganize tests into separate modules

nfahlgren:
**Describe your changes**
Reorganizes tests into individual modules instead of one large tests file. Streamlines/reorganizes test data to remove unnecessary/redundant test data in favor of smaller/faster test samples. Updates tests to use parameters and test a single function per test.

**Type of update**


nfahlgren:
@HaleySchuhl @JorgeGtz this PR now re-implements our tests with 100% coverage but split into smaller modules, feel free to take a look when you have time. There are a lot of files changed but most of those changes are:

* Addition of new test module files
* Moving data from the original tests data directories to the unified `testdata` directory
* Removal of test data that is no longer used/needed

There are a few minor updates to some PlantCV functions that should just refactor the code to remove the need for a few tests (e.g. removed all cases of `if params.debug is not None`).

So other than these few latter cases, most of the other file changes don't need to be reviewed thoroughly since they are just cleaned up tests moved from `tests.py` into modules.

One last thing, the empty file `tests.py` is still included because I think that will help with conflict resolution as we merge these changes into existing PRs. We can remove it later after we've dealt with those conflicts.

HaleySchuhl (approved):

PR #856, opened 2022-01-20
add line return to Sample Images function

### LINKED ISSUE #855: sample_images function does not output line returns
**Describe the bug**
`pcv.utils.sample_images` on line 42 does not include a line return character, so all rows in `SnapshotInfo.csv` are written out on one line. The code should be:

```python
out_file.write(','.join(row) + "\n")
```

HaleySchuhl:
add in a line return to the step where new snapshot info CSVs are getting printed out

**Type of update**
Is this a:
* Bug fix

**Associated issues**
- closes #855 

**Additional context**


nfahlgren (approved):
Thanks @HaleySchuhl!

PR #864, opened 2022-02-18
Update versioneer from 0.18 to 0.21

nfahlgren:
**Describe your changes**
Updates `versioneer` from 0.18 to 0.21.

**Type of update**
Is this a: update


nfahlgren:
I think the missing coverage for `_version.py` will be fixed once merged. We ignore coverage of it since it's a file installed by `versioneer`. They introduced new functions that I think are not being ignored because the main branch doesn't have the new `.coveragerc` configuration that's introduced in this PR

nfahlgren:
@HaleySchuhl @JorgeGtz

Issue #881, opened 2022-03-29
JSON to csv


stiphyMT:
**Describe the bug**
Usually I run a workflow with conditionals flow for different angle or camera images. This creates NA values for certain variables that do not exist is a certain angle/camera image combination. For conversion of the multiple-value variable data to csv, in the resulting file the cases with the NA have one less column, the _sample_ variable/column is missing.

I found this on PlantCV v3.13 but looking at the github history it has existed since v3.11

I believe that in the file 
**/plantcv/tree/master/plantcv/utils/converters.py**, line **76**
`csv.write(",".join(map(str, meta_row + [var, "NA", "NA"])) + "\n")`
should be
`csv.write(",".join(map(str, meta_row + [sample, var, "NA", "NA"])) + "\n")`

thus aligning the columns with the non exception case on line **73**
`row = [sample, var, obs[var]["value"][i], obs[var]["label"][i]]`

Stephan

stiphyMT:
PS I haven't used GitHub much interacting with others until just now. How do you change the Labels? I want 'bug' and not 'help wanted'.

HaleySchuhl:
Hi @stiphyMT ,

Thank you for opening an issue and bringing this to our attention. I added the `bug` tag to this issue; in GitHub the labels are located along the right hand side of an issue once it's been posted. 

Would you be interested in being added as a contributor of PlantCV so you'd be able to open Pull Requests to update and/or add code? Totally optional!

stiphyMT:
Ciao Haley,
If you are patient with me I can give it a try. While I use git, when it comes to github I am never sure what I am doing is what I want. Should I make a pull request for this issue? (something small that I can start with).
Stephan

PR #882, opened 2022-04-02
added sample column to case without variables

stiphyMT:
**Describe your changes**
When converting JSON file to csv, the multi-trait lines that do no have values (i.e. NA) are missing the 'sample' column/value (line 76) as compared to the normal case (line 73). 
`csv.write(",".join(map(str, meta_row + [var, "NA", "NA"])) + "\n")`
The 'sample' value was added to the list on line 73. 
`csv.write(",".join(map(str, meta_row + [sample, var, "NA", "NA"])) + "\n")`

**Type of update**
Is this a:
* Bug fix

**Associated issues**
This change closes issue #881.

**Additional context**
Before this modification the resulting csv file could result with some rows having one too few columns, resulting in errors during import (at least this was the case using fread of the data.table library from R). With this change everything appears normal.

### DIFF
--- a/plantcv/utils/converters.py
+++ b/plantcv/utils/converters.py
@@ -73,7 +73,7 @@ def json2csv(json_file, csv_file):
                                 row = [sample, var, obs[var]["value"][i], obs[var]["label"][i]]
                                 csv.write(",".join(map(str, meta_row + row)) + "\n")
                     else:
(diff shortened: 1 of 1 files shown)


nfahlgren (approved):
This looks great, thanks @stiphyMT!

Issue #883, opened 2022-04-03
More color depth support needed


Clarivy:
Hi! 
I'm using plantcv.transform.correct_color to match two color checkers. But I'm having some issues. 
The images in my project are in 16-bit color depth which is not supported according to your source code. The outcome images are incorrect.

I checked the source code. And I found some snippets like this:

```python
    # convert img to float to avoid integer overflow, normalize between 0-1
    source_img = source_img.astype(np.float64)/255
```

To support deeper color depth, you should check whether img.dtype is numpy.uint8, numpy.uint16 or numpy.float32. And divide img by different value.

nfahlgren:
Hi @Clarivy, @JorgeGtz has worked on implementing this in the color correction module in #885. It works but we realized that there are going to be bigger issues if you were planning on using the corrected images with other PlantCV modules.

The two main issues we know about are 1) that `matplotlib`, which we use for displaying images, clips 16-bit data to 8-bit, so the images are not displayed correctly; and 2) that OpenCV does not support 16-bit color images.

A couple things we discussed:
* If you just need color correction and do not plan on using other PlantCV functions, then I think we could accept the changes in #885 because they wouldn't affect our standard functionality but would allow you to use color correction on your data.
* If you were planning on utilizing PlantCV, then I think your data would need to be converted to 8-bit. We discussed the potential need for a new function that would correctly scale the data from one data type to another that could be useful in this case.

Please let us know your thoughts, thanks!

Clarivy:
Thank you for your rapid reply.

I am only using the color correction module currently. Higher precision is really important to me, so the implementation is really helpful :)

nfahlgren:
Thanks @Clarivy, PR #885 introduces these changes and we just merged them into the main branch.

nfahlgren:
We also added this update to the release v3.14, which is now available on PyPI. It should also be available through conda within a day or so.

PR #885, opened 2022-04-04
support for other dtypes in color correction

JorgeGtz:
**Describe your changes**
Modify the color correction functions so they work with all unsigned integer and float  dtypes (assuming images are between 0-1 for floats).
The normalization to work with values between 0-1 is done depending on the dtype and the corrected image has the same dtype as the source image.

**Type of update**
Is this a:
* Bug fix

**Associated issues**
#883 

**Additional context**
Before this PR the source and target images were normalized by dividing by 255 which is only valid for uint8 dtype

### DIFF
--- a/plantcv/plantcv/transform/color_correction.py
+++ b/plantcv/plantcv/transform/color_correction.py
@@ -236,18 +236,18 @@ def apply_transformation_matrix(source_img, target_img, transformation_matrix):
     # merge corrected color channels onto source_image
     bgr = [b, g, r]
     corrected_img = cv2.merge(bgr)
+    print(corrected_img.dtype)
 
     # return values of the image to the original range
     corrected_img = max_val*np.clip(corrected_img, 0, 1)
-    corrected_img = np.floor(corrected_img)
-    # cast back to original dtype
+    # cast back to original dtype (if uint the value defaults to the closest smaller integer)
     corrected_img = corrected_img.astype(source_dtype)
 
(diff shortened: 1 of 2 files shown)


plantcv/plantcv/transform/color_correction.py
```diff
@@ -35,8 +35,15 @@ def get_color_matrix(rgb_img, mask):
     if len(np.shape(mask)) != 2:
         fatal_error("Input mask is not an gray-scale image.")
 
+    img_dtype = rgb_img.dtype
+    # normalization value as max number if the type is unsigned int
+    if img_dtype.kind == 'u':
```
nfahlgren, line 40: One way to simplify this so you don't have to increase the number of tests is:

```
max_val = 1.0
if img_dtype.kind == 'u':
    max_val = np.iinfo(img_dtype).max
```

plantcv/plantcv/transform/color_correction.py
```diff
@@ -203,8 +210,14 @@ def apply_transformation_matrix(source_img, target_img, transformation_matrix):
     # split transformation_matrix
     red, green, blue, red2, green2, blue2, red3, green3, blue3 = np.split(transformation_matrix, 9, 1)
 
+    source_dtype = source_img.dtype
+    # normalization value as max number if the type is unsigned int
+    if source_dtype.kind == 'u':
```
nfahlgren, line 215: Same as above

HaleySchuhl (approved):

PR #887, opened 2022-04-26
Refactor parallel subpackage to use dataframes and grouping

### LINKED ISSUE #423: parallel pipeline for image series

### LINKED ISSUE #474: Store timestamps in ISO 8601 format
Currently, timestamps are not transformed, they are stored in the format they are found in. Datetime strings are converted to datetime objects when filtering by a date range is supplied but are otherwise ignored.

Storing datetime strings in a common format would make the timestamps more useful to downstream analysis so that datetime parsing doesn't need to be done later. In #473 we should add the ability to parse a wide range of datetime formats, which would enable us to then process all datetime strings into ISO 8601 formatted strings for consistency.

We should also consider converting timestamps to UTC but would need to deal with locale.

**Additional context**
ISO 8601: https://en.wikipedia.org/wiki/ISO_8601

### LINKED ISSUE #538: Overall improvements to workflow parallelization
This is an Epic to help organize the overall efforts to improve PlantCV workflow parallelization. There are three main components to workflow parallelization:

* Metadata processing/handling
* Workflow handling
* Parallelization

Various improvements have been suggested for both areas. We will attempt to subdivide these into discreet issues as much as possible, but many improvements will be inherently interlinked. Areas of improvement are (in no particular order):

- [x] Update input parameter names for consistency #470
- [x] Utilize configuration files instead of lengthy command-line options #470
- [x] Image grouping or aggregation for data series (e.g. multi-frame image sets, co-processing, etc.) #423
- [ ] Adding metadata directly in workflows #490
- [ ] Improve/expand valid metadata terms
- [x] Allow multiple metadata terms in the metadata filter #385 #461
- [x] New batch processing system(s) #28
- [x] Store timestamps in ISO 8601 format #474

Will update if I missed anything or plans change.

nfahlgren:
**Describe your changes**
This PR makes several major changes to the parallel subpackage.

`.github/workflows/continuous-integration.yml`:
* Use `pip` to install PlantCV instead of `setup.py`.

`plantcv/parallel/__init__.py`:
* Remove deprecated functions `convert_datetime_to_unixtime` and `check_date_range`.
* Import new `workflow_inputs` function and `WorkflowInputs` class. These are from new module for handling workflow inputs in Jupyter notebooks and scripts was added to make it easier to migrate from Jupyter to a parallel workflow script.
* `WorkflowConfig` default timestamp format was updated to an ISO 8601 UTC datetime.
* Remove `WorkflowConfig` `coprocess` attribute and replace with `groupby` and `group_name` attributes. The new attributes are used to group images in the new dataframe-based metadata parser framework and name the image inputs to parallel workflows.
* A new `rotation` metadata attribute was added to `WorkflowConfig`.
* The `plantbarcode` metadata attribute in `WorkflowConfig` was renamed to `barcode` to be more general.

`plantcv/parallel/parsers.py`:
* The parsers module was completely rewritten to reduce complexity and utilize dataframes for easier filtering and grouping.
* A new input data format (`phenodata`) was added to the parsers module.
* All data formats in the parsers module are read into a common data structure, which allows them to all be converted to a dataframe using a single method. New data formats can be added by adding a new private function and reading the data into the common data structure.
* Dataframes in the parser module can be grouped on one or more metadata terms, the grouped dataframe is passed to the job builder module.

`plantcv/parallel/job_builder.py`:
* The job builder module was updated to handle grouped dataframes and remove coprocessing logic.
* The job builder module was updated to pass arguments to the new `workflow_inputs`-based `argparse` framework.
* The job builder module can automatically name workflow input images if requested (e.g., `image1`, `image2`, etc.)
* Temp JSON files are now named with unique alphanumeric identifiers (UUID) since they can represent data for multiple images.

`plantcv/parallel/workflow_inputs.py`:
* New module for defining/handling standard workflow inputs in both Jupyter and parallel workflow script contexts.
* The `WorkflowInputs` class is used to set Jupyter notebook input variables in a framework that is compatible with the command-line arguments used in parallel workflow scripts.
* The `workflow_inputs` function creates a standardized `argparse` command-line argument parser for workflows.

`plantcv/parallel/process_results.py`:
* The JSON output file is now formatted with line returns and indentation for easier viewing.

`plantcv/utils/converters.py`:
* The `json2csv` util function was updated to handle grouped output data.
* `json2csv` now only outputs a single CSV file in long format.

Additionally, relevant tests were added/updated. Documentation was updated where necessary.

**Type of update**

**Associated issues**
Closes #474
Closes #423 
Closes #538
Replaces #759


HaleySchuhl:
Error handling for "images" input in Workflow Image when input is not a list.

PR #998, opened 2023-03-03
Fix ImportError in plantcv caused by change in skimage.feature function name

fbarbe00:
Currently, when installing plantcv, the following error is raised: `ImportError: cannot import name 'greycomatrix' from 'skimage.feature'`
This is because skimage.feature changed their function name from greycomatrix to graycomatrix (see https://stackoverflow.com/questions/67537650/cannot-import-name-graycomatrix-from-skimage-feature ) from version 0.19. This commit changes the requirements.txt file and the function name.

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -16,7 +16,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        python-version: [3.7, 3.8, 3.9]
+        python-version: ['3.8', '3.9', '3.10']
         os: [ubuntu-latest]
     env:
       OS: ${{ matrix.os }}
--- a/.github/workflows/publish-package.yml
+++ b/.github/workflows/publish-package.yml
@@ -12,7 +12,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        python-version: [3.9]
+        python-version: ['3.10']
         os: [ubuntu-latest, macos-latest, windows-latest]
     steps:
     - uses: actions/checkout@main
--- a/environment.yml
+++ b/environment.yml
@@ -2,7 +2,7 @@
 # optionally, change channel name with -n {plantcv-dev}
 name: plantcv
 dependencies:
-  - python=3.9
+  - python=3.10
   - matplotlib>=1.5,<3.6
   - numpy>=1.11,<1.23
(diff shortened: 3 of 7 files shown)


HaleySchuhl:
@fbarbe00 thanks very much for opening this PR and sorry for the delay in response.

HaleySchuhl:
Ugh..Looks like the renaming issues also trickle down past the imports part. 

Now tests fail on 
- `plantcv/plantcv/closing.py:26: in closing` -->  `E       TypeError: binary_closing() got an unexpected keyword argument 'selem'`
-  The `watershed_segmentation ` function --> `E       TypeError: peak_local_max() got an unexpected keyword argument 'indices'` 

If you'd prefer, I can dig into their updated parameter names, push commits to this PR, and hopefully resolve those remaining bugs. Thanks again @fbarbe00 for bringing this to our attention. @nfahlgren can we add them as a contributor so tests automatically run next time?

fbarbe00:
Thanks a lot for your reply! Is there anything I can still try to help with?

fbarbe00:
Btw, this solves Issue #1002

HaleySchuhl:
Yes @fbarbe00 If you're able to figure out how to get an array-like output from the latest version of `skimage.feature.peak_local_maxskimage.feature.peak_local_max` .  TThe that functionality isn't recoverable I might suggest pinning the version down to the older functionality.

fbarbe00:
Hi @HaleySchuhl , I believe my last commit fixes it! I just looked at what the `peak_local_max` function did in `v0.18` (https://github.com/scikit-image/scikit-image/blob/v0.18.x/skimage/feature/peak.py) , and looks like `indices=False` only affected two lines at the end. These lines have now been added to plantcv.

Note that I do not have a deep understanding of what this code actually does, but this does seem to fix this function and work as before.

HaleySchuhl:
Awesome! Tests seem to pass but now there's an issue in the Python 3.7 build on codecov for some reason? `ErrorDetail(string='Unable to locate build via Github Actions API. Please upload with the Codecov repository upload token to resolve issue.', code='not_found')`

nfahlgren:
I think codecov stopped working with Python 3.7, which was the oldest version we were testing. I bumped up the testing environments to 3.8, 3.9, and 3.10 since 3.11 is out now.

HaleySchuhl (approved):

Issue #978, opened 2022-11-08
Typo in read_cropreporter example


wurDevTim:
**Additional context**
Using plantcv 13.4.1 on a windows 10 machine. 

**Describe the bug**
In the documentation: https://plantcv.readthedocs.io/en/4.x/photosynthesis_read_cropreporter/
The function is called as follows:  
`ps = pcv.photosynthesis.read_cropreporter(filename="PSII_HDR_020321_WT_TOP_1.INF")`

If we than look in the function we see it expects to be called with the '.DAT' file, line 28,29 of  https://github.com/danforthcenter/plantcv/blob/main/plantcv/plantcv/photosynthesis/read_cropreporter.py

```
    inf_filename = filename.replace("PSD", "HDR")
    inf_filename = inf_filename.replace(".DAT", ".INF")
```

**Expected behavior**
- The example is updated to call the function with a '.DAT' file to avoid confusion:
  `ps = pcv.photosynthesis.read_cropreporter(filename="PSII_PSD_020321_WT_TOP_1.DAT")`
- Our cropreporter also outputs 'CHL', which contain a similar content. It would be nice if you either add support for this (new feature) or add a note that only 'PSD' files are supported.

HaleySchuhl:
Thanks for bringing this to our attention @wurDevTim ! 

I was curious if this typo was showed up anywhere else, and it looks like the 4.x branch has this typo two places (also [here](https://plantcv.readthedocs.io/en/4.x/tutorials/psII_tutorial/)) `./workflowname.py -i PSII_HDR_020321_WT_TOP_1.INF -o /home/user/output-images -D 'print'`.  The rest of the tutorial looks fine, but I think it's about time we do an annual comb through for any updates that didn't get propagated down into tutorials. Please let us know if you come across anything else like this!

nfahlgren:
I don't think there is a typo. In PlantCV v3 the input file is the DAT file, but in v4 it is the INF file. In the links above you are comparing the documentation from 4.x to the code in the `main` branch, which is v3. We had to make this change because in v3 we only supported reading the PSD measurement file but in v4 we can read PSD, PSL, CLR, CHL, etc.

In the 4.x branch we have been developing a v4 release, and in v4 the `photosynthesis` submodule was completely redeveloped and supports more of the measurement routines from CropReporter systems. Versus in v3 we only supported Fv/Fm.

[...]

PR #1217, opened 2023-07-06
Cropreporter adding pam

wurDevTim:
As mentioned in https://github.com/danforthcenter/plantcv/issues/978 we first wanted to add OJIP and than follow with other protocols. At the moment a lot of researchers are using PAM, there where some issues with the format. Luckily a colleague at the WUR managed to figure this out. I will double check with colleagues but for as far as I know the PSII efficiency calculation are the same for OJIP and PAM. 

The adaptation takes quite some time, therefor we quickly collected this test data by running the different protocols after each other. I will request a measurement day to collect data from a real plant. 

There are a lot of big changes in the 'release-4.0' branch which take some time to process, please let me know if I missed something. 

**Type of update**
New feature or feature enhancement

**Associated issues**
[Reference associated issue numbers. Does this pull request close any issues?](https://github.com/danforthcenter/plantcv/issues/978)

**Additional context**
Upgrading the photosynthesis library to support the new capabilities used at WUR. 

**For the reviewer**
- [ ] All tests pass
- [ ] Documentation tested
- [ ] Code reviewed
- [ ] PR approved


nfahlgren:
Hi @wurDevTim, thanks for the further updates! We have been doing a little work to understand the differences between the measurement protocols just so that we can make it clear in a tutorial what they all mean.

One question directly about this pull request. The function `pcv.visualize.chlorophyll_fluorescence` was designed to plot the fluorescence induction curves produced by the OJIP method. Since the PAM method does not produce the same curve, should we have the function recognize the PAM datasets still? If no, is a different plot useful? @kmurphy61 and I were discussing adding an additional time value frame label to the data structure so that we could potentially plot something like this: https://www.researchgate.net/profile/Alexander-Ruban-2/publication/319123856/figure/fig2/AS:631664216653865@1527611893716/Typical-PAM-fluorescence-measurement-of-an-Arabidopsis-leaf-showing-induction-and_W640.jpg

Some other things we discussed, that could probably be a separate pull request after merging this, are:
* Renaming the datasets because `darkadapted`, etc. are no longer specific. We could instead have something like `ojip_dark`, `ojip_light`, `pam_dark`, `pam_light`.
* Are the PAM data compatible with measuring NPQ? We think not based on the documentation but are not 100% sure.

nfahlgren:
@all-contributors please add @wurDevTim for code, doc, test.

wurDevTim:
Heey @nfahlgren, due to vacations things are moving a bit slower these months. 
Regarding plots, the fluorescence induction curves are most interesting for the OJIP protocol when all frames are saved. If we don’t save all frames or do a PAM measurement we get a fixed number of frames and this graph does not say much. The graph you suggest could be nice but I don’t think we can get enough information from the cropreporter to create it.  Here I have to admit that I am no photosynthesis expert, luckily some of my colleagues are. Usually they don’t even look at the fluorescence, instead they directly compute the NPQ and PSI2. Therefore I doubt if it is worth changing now. 

You might have noticed that the fluorescence and photosynthesis look terrible on the test data, this is due to the lack of adaption on the images of this leaf. Usually we don’t run all protocols on the same plant, but there is data coming. When that is in we plan to dive into the NPQ & PSI2 calculations. 
If you want a more insight in how and what is typically done at the WUR I can try to setup a meeting? For the more technical question we have the option to call in support from phenovation. 

Lastly, if the renaming does not break backwards compatibility it would be nice, I can pick that up next.

HaleySchuhl:
Can you ping me when this is ready to review? 💡

wurDevTim:
Hello @nfahlgren, 
It has been an interesting few days here, first we collected new test data which looked good on first sight. 
However, in the `pcv.visualize.chlorophyll_fluorescence` plot of the light adapted plant, added below, my colleague saw that the Fmp has a lower value than the last frame. This means the lighting used to take the Fmp was not saturating. 
image
We will be scanning to many plants to analyze all these plots, but it could be used to get a fast overview. 

Not sure why, but phenovation called the PAM variant of NPQ is called the `PAM time protocol`. 
image
Both the OJIP & PAM variant generate special filetypes on our cropreporter. 
We have not tried these modes yet, I will ask if it's useful to add them already.  

Regarding the extra plots, we rarely look at individual plants. Usually we track dozens of plants over time, after computing the PSII or NPQ some statistical correction is applied. All these measurements are than plotted in a single graph to check for differences between treatments. 

There are a few more parameters which are often computed:
- Linear electron flow
- Energy-dependent quenching (Qe)
- Photo inhibitory quenching (Qi)
These are a bit debated so I am sure if it's wise to add them to the repository? 
Its also very simple to compute these parameters and many more ourselves when we have the frames in the `PSII_data class`.

nfahlgren:
Hi @wurDevTim, we could definitely circle back around to the PAM time protocol data. Some folks here are interested in Qe, etc., so worth adding if you also think it's useful for what you're doing. Thanks!

nfahlgren (approved):

Issue #886, opened 2022-04-18
Save results in CSV format without overwriting the file


nfahlgren:
**Is your feature request related to a problem? Please describe.**
`plantcv.outputs.save_results` will add data to an existing JSON file but it overwrites an existing CSV file. If you want to combine results from multiple PlantCV workflows, from multiple Jupyter notebooks for example, you currently need to save multiple CSV files and then combine them (@josectovar brought this to our attention).

**Describe the solution you'd like**
We could open the CSV file with mode="a" (append) instead of overwrite. We would need to exclude the header if the file exists to avoid duplicating it when adding results. We would have to assume that new observations are named uniquely relative to data already in the output file.

**Describe alternatives you've considered**
We could potentially read the existing data as a dataframe and append the new data as a dataframe and save the merged dataframe as CSV.

afinit:
Came up with one solution in #1225 but it needs some discussion

PR #1225, opened 2023-07-08
Release 4.0 append output for save_results

afinit:
## Describe your changes
I expect changes, but this seems like a good start for discussion.

- adds error handling for an illegal `outformat` param
- adds append param that works for both json and csv. defaults to `True`
- changes the code to use `with` to open csv file

Before this PR, neither json nor csv files would append data. The json path reads in the existing data, but throws away the observations and only keeps the metadata.

The part that needs some discussion:
With this PR, existing json and csv data are treated differently. When json data is appended, we overwrite any observations that have the same sample name as our new data. When csv data is appended, we are strictly appending without any consideration of the data that is already in the file.

I don't like this difference in behavior

### Options for fixing the differences

There are probably other options, but these are the 2 ideas I came up with to resolve the current difference in behavior

#### Option 1
Read existing CSV data with pandas and remove any rows with the same sample name as the new data

#### Option 2
Read existing data for both json and csv. Modify conflicting sample names to have an appended `_###` where we autoincrement that number for new data. This would look something like:
```
# existing data:
sample,trait,value,label
default,string,string,none
default_001,boolean,1,none

# new data:
sample,trait,value,label
default,string,string2,none
default,boolean,0,none

# data written:
sample,trait,value,label
default,string,string,none
default_001,boolean,1,none
default_002,string,string2,none
default_002,boolean,0,none
```

## Type of update
new feature

## Associated issues
Addresses #886 

## For the reviewer
- [ ] All tests pass
- [ ] Documentation tested
- [ ] Code reviewed
- [ ] PR approved


nfahlgren:
We ended up deemphasizing CSV outputs because JSON temp files fit into other parts of our toolset better and we can combine and convert these to CSV with existing tools

Issue #1481, opened 2024-03-21
Watershed debug image getting returned


HaleySchuhl:
**Describe the bug**
Running `pcv.watershed_segmentation` ought to return an output image that might be useful for downstream analysis. Currently, the function returns this visually helpful (but analytically weird) `joined` image when really it should probably return the `labels` output from line 55 where we run the scikit image watershed function. This bugfix was caught by @maliagehan and implemented into [this branch](https://github.com/danforthcenter/plantcv/pull/754/files#diff-62b214ca9d73cf0cbd5fc1bea180ff2fa4bf8c22e5a8c72cb788a5b0e586e6f9) but we have since separated out the changes in this PR for smaller, more manageable PRs. I will open a new PR shortly to address this. 

**To Reproduce**
Run `pcv.watershed` and then plot the returned image 
![image](https://github.com/danforthcenter/plantcv/assets/44006936/e272d47c-ff2c-4ecc-b6c5-17523885a92d)

**Expected behavior**
Grayscale labeled mask for output 

**Local environment (please complete the following information):**
 - OS: macOS
 - Environment: conda
 - PlantCV Version 4.2.1+264.gcc1e8725

**Additional context**

PR #1482, opened 2024-03-21
Update watershed.py to make the function return `labels` 

HaleySchuhl:
For context please see... 
**Associated issues**
- #1481 

**For the reviewer**
- [x] All tests pass
- [x] Documentation tested
- [x] Code reviewed
- [x] PR approved


plantcv/plantcv/watershed.py
```diff
@@ -86,6 +86,6 @@ def watershed_segmentation(rgb_img, mask, distance=10, label=None):
                             value=estimated_object_count, label='none')
 
     # Store images
-    outputs.images.append([dist_transform, joined])
+    outputs.images.append([labels, joined])
```
nfahlgren, line 89: We can remove this code, we don't use the stored images anymore

plantcv/plantcv/watershed.py
```diff
@@ -75,8 +75,8 @@ def watershed_segmentation(rgb_img, mask, distance=10, label=None):
 
     # Reset color sequence mode
     params.color_sequence = color_sequence
-    _debug(visual=dist_transform,
-           filename=os.path.join(params.debug_outdir, str(params.device) + '_watershed_dist_img.png'),
+    _debug(visual=labels,
+           filename=os.path.join(params.debug_outdir, str(params.device) + '_watershed_labels_img.png'),
            cmap='gray')
     _debug(visual=joined,
            filename=os.path.join(params.debug_outdir, str(params.device) + '_watershed_img.png'))
```
nfahlgren, line 82: I think we should remove this debug image and instead should use [label2rgb ](https://scikit-image.org/docs/stable/api/skimage.color.html#skimage.color.label2rgb) to colorize the labeled mask for debug (rather than plotting the grayscale labels - like we do in `pcv.create_labels`).

nfahlgren (changes requested):

plantcv/plantcv/watershed.py
HaleySchuhl: Good idea! I can implement these

HaleySchuhl:
Updated debug image so will update example in docs page tomorrow probably

HaleySchuhl:
> Just one comment before approving: I don't think the documentation makes it clear how the labels are/can be used or accessed. For example, it says "These data can always get accessed during a workflow (example below)" but the labels are not accessed in the example. Maybe this is better demonstrated by a tutorial, but I think a small example in the documentation page could help illustrate the functionality.

Good point, just added~~ Thanks Anna!

annacasto (approved):

nfahlgren (approved):

PR #1518, opened 2024-05-09
bugfix for roi_type="largest" in roi.filter

### LINKED ISSUE #1507: Bug in logic behind `roi_type="largest"` option inside `pcv.roi.filter`
**Describe the bug**
A bug was found in some applications of `roi_type="largest"` within the `pcv.roi.filter` function recently, uncovered by @sdkenney42 . Instead of returning the largest object in the mask it was consistently returning the second largest object for a specific image. We did a little digging into the function at lab meeting. The logic behind this function is unsound because it is checking the length of a contour rather than actually filtering on contour size. One option for updating `_helpers.py` instead of checking the length at line 96 we should actually judge by contour area. 

**To Reproduce**
```
%matplotlib widget 

import numpy as np 
import cv2

# Make a blank mask
mask = np.zeros((1000,1000), dtype=np.uint8)
start_point = (10, 10) 
end_point = (110,110) 
# Draw 100x100 pixel square 
mask = cv2.rectangle(mask, start_point, end_point, 255, -1) 
#pcv.plot_image(mask) 

# Calculate radius (based on h=100 in our square) that will give nearly equal areas 
circle_r = int(100/np.sqrt(np.pi)) + 1
#print(circle_r)
mask = cv2.circle(mask, (500,500), circle_r, 255, -1) # circle with radius of 3

roi = pcv.roi.rectangle(img=mask, x=1, y=1, h=49, w=49)

pcv.params.debug = "plot" 
filtered_mask = pcv.roi.filter(mask=mask, roi=roi, roi_type='largest')  # Square gets kept 
vis = pcv.visualize.obj_sizes(img=mask, mask=mask)  # But the area of the circle is larger 
``` 

**Expected behavior**

 - OS: [macOS]
 - Environment [conda.]
 - PlantCV Version [main branch 4.2.2.dev432+g8844cfc8.d20240426 ]

**Additional context**
I believe `len(contour)` is more equivalent to taking the perimeter vs the area. Instead need to judge which contour to keep with more sound logic, perhaps the solution below. 
```
c = max(contours, key = cv2.contourArea)
```
or ID the largest contour after filtering on ROI, and then `pcv.fill` anything smaller than the known largest object size.

HaleySchuhl:
**Describe your changes**
Updates the way we determine which contour is largest after filtering in `pcv.roi.filter` with `roi_type="largest"` setting. Merge this AFTER #1520 

**Type of update**
Is this a:
* Bug fix
* Work in progress

**Associated issues**
- closes #1507 
- 
**Additional context**

**For the reviewer**
- [x] All tests pass
- [x] Documentation tested
- [x] Code reviewed
- [x] PR approved


HaleySchuhl:
Added different assertions for the unit test. Instead of saying the contour is exactly `xx` pixels, we can make sure that the contour size is the same as before filtering out the small object. Also added a small object to the "pre" mask so that the function is actually doing something.

nfahlgren (approved):

PR #1576, opened 2024-08-07
Add "Quick Cutto" ROI function

HaleySchuhl:
**Describe your changes**
A much faster method to making filtered binary and labeled masks from one or more ROIs. Plan to merge this functionality into other plantcv functions in further PRs. 

**Type of update**
Is this a:
*  feature enhancement

**Associated issues**
Reference associated issue numbers. Does this pull request close any issues?

**Additional context**
The iterative nature of `pcv.roi.filter` and therefore also `pcv.create_labels` makes the function speed a bottleneck when image size, mask complexity, and the number of objects/ROIs gets larger.  

**For the reviewer**
- [x] All tests pass
- [x] Documentation tested
- [x] Code reviewed
- [x] PR approved


nfahlgren:
One quick thing, does that maize image exist already? If so we don't need to include a copy of it in a separate docs folder

HaleySchuhl:
> One quick thing, does that maize image exist already? If so we don't need to include a copy of it in a separate docs folder

I was looking for it but I think it might only exist in the tutorial repo(s). That reminds of something we talked about at dev meeting last week though, about ways to make downloading the repo smaller. Not sure if it would make a significant/any difference relevant to the effort it would require, but there are definitely replicate doc images that we could try to consolidate. Maybe the most glaring example of the duplication I've come across while looking for the maize image is the "original_img" in each of the `pcv.rgb2__` doc pages.

nfahlgren:
Thanks Haley, you're right that the more images there are (docs and tests) the bigger the repo gets and the longer downloads take. Fortunately that only impacts folks installing from source code, the tests and docs are not included in the PyPI and conda packages

k034b363 (approved):
Looks good to me! I tested it with circle, rectangle, custom, and multi ROI objects and it works just fine with all of them. I removed the docs page, but before it goes back in it'll have to be reviewed again to make sure the output names match the function docstring, but that was all I saw that I would've suggested changing.

PR #1593, opened 2024-09-03
Added NPCI and GLI indices to spectral_index

rayn-alex:
**Describe your changes**
I added two additional indices, Normalized Pigment Chlorophyll Index and Green Leave Index, to `spectral_index`. Added also the respective tests and documentation (including references).

**Type of update**
Is this a:
* Update to documentation

**Associated issues**
None

**Additional context**
None

**For the reviewer**
- [x] All tests pass
- [x] Documentation tested
- [x] Code reviewed
- [x] PR approved


HaleySchuhl:
Thanks @rayn-alex for adding this! I'll plan to review it early next week.

I have also asked @nfahlgren to add you as a contributor so that you can branch the main plantcv repo in any future PRs (DeepSource is not reporting code coverage since it's branched from your repo). Hope you have a great rest of your Friday.

rayn-alex:
Great, thank you and good to know! Already received the invitation.

Have a pleasant week ahead!

HaleySchuhl:
Good morning @rayn-alex , thanks again for contributing.

I've started on reviewing this PR and the `NPCI` functionality & docs look great. 

I noticed `GLI` can be calculated from the bands available in an RGB image, so to match the patter in other indices like this (e.g. EGI) I believe we should update the input parameter from `hsi` to `rgb_img`. While testing each new index on the small maize HSI that's used in tests, I noticed the resulting GLI index has a max value of 183 although the predicted data range for the index is [-1,1]. This prompted me to check the [publication of the GLI index](https://www.researchgate.net/profile/Mounir-Louhaichi/publication/228906661_Spatially_Located_Platform_and_Aerial_Photography_for_Documentation_of_Grazing_Impacts_on_Wheat/links/53ee2a8a0cf23733e80b3265/Spatially-Located-Platform-and-Aerial-Photography-for-Documentation-of-Grazing-Impacts-on-Wheat.pdf) and they specify the input bands ought to be [0,255], which is true of the pseudo_rgb. I tested this with the code below. 
```
%matplotlib widget
import os
import cv2
import matplotlib
import numpy as np

pcv.params.debug = "plot"
test_img = "../../GitHub/plantcv/tests/testdata/corn-kernel-hyperspectral.raw"
# Read in the hyperspectral data
array = pcv.readimage(filename=test_img, mode="envi")

rgb_img = array.pseudo_rgb

blue, green, red = cv2.split(rgb_img)
# Calculate GLI 
## GLI = (2 * GREEN - RED - BLUE) / (2 * GREEN + RED + BLUE)
with np.errstate(divide="ignore", invalid="ignore"):
    r = red.astype(np.float32)
    g = green.astype(np.float32)
    b = blue.astype(np.float32)
    index_array_raw = (2 * g - r - b) / (2 * g + r + b)

print(np.max(index_array_raw))
print(np.min(index_array_raw))
```

If it's ok with you, I can make these changes today, and then have you test them locally before Noah merges the changes into main?

rayn-alex:
Hey @HaleySchuhl,

I added the GLI index as `hsi` index mainly because I'm currently only working with hyperspectral images and going the through `pseudo_rgb` would change the compatibility.  

I absolutely see the point and I think it should be changed as you suggested it. 

However, would it make sense to add support for both input types, `rgb_img` and `hsi`? It's probably just a simple input check in the beginning. I could take care of it. I think it would only affect EGI and the GLI index so far.

What do you think?

HaleySchuhl:
@rayn-alex I really like the idea of added flexibility to accept both HSI and RGB in those two indices. Very in favor of detecting and handling expected datatypes. 

The one consideration is that changing the input parameter in EGI to something more general like `img` would be a breaking change, and we're trying not to introduce any of those between major releases. So can you go ahead and add the check to both functions and allow them to handle both input datatypes, but preserve the `rgb_img` input name in EGI?

rayn-alex:
@HaleySchuhl 
Great! Sure, I can preserve the `rgb_img` in EGI. I'll work on this in the next days.

rayn-alex:
@HaleySchuhl Done! 

I added new tests and updated the docs as well. I used different wavelength presets for the EGI RGB wavelength and also increased the default distance to 40nm because the original publication only uses the RGB values. 

~R700 for RED, ~R530 for GREEN, and ~R460 for BLUE

HaleySchuhl:
Happy Monday @rayn-alex ,

I'm reviewing the updated functionality this morning and pushed a few minor changes so far related to DeepSource complaints. 


While testing the functions locally, I ran into similar inconsistent-data-range outputs for the GLI. The RGB input functionality works nicely and gives an index with values all within the predicted range [-1,1]. However, the `Spectral_data` case still is not rescaling the reflectance bands to int8/[0,255] prior to calculating the index, so the calculated GLI on the unit test image still has a max value far outside the predicted range. 


Maybe re-scaling the HSI bands is less than ideal, so I'm open to other solutions that seem appropriate.

plantcv/plantcv/spectral_index/spectral_index.py
```diff
@@ -458,6 +523,37 @@ def ndre(hsi, distance=20):
...
+
+    :param hsi: __main__.Spectral_data
+    :param distance: int
+    :return index_array: __main__.Spectral_data
+    """
+    if (float(hsi.max_wavelength) + distance) >= 680 and (float(hsi.min_wavelength) - distance) <= 430:
+        r430_index = _find_closest(np.array([float(i) for i in hsi.wavelength_dict.keys()]), 430)
+        r680_index = _find_closest(np.array([float(i) for i in hsi.wavelength_dict.keys()]), 680)
+        r680 = (hsi.array_data[:, :, r680_index])
+        r430 = (hsi.array_data[:, :, r430_index])
```
nfahlgren, line 548: Should these be cast to float32 also?

rayn-alex: When I introduced this index, I was copying the structure of others. As far as I can see, there is no cast to float32 in other indices at this step. 

But since you brought this up, it might make sense to introduce a cast to float, possibly already in `read_data` depending on the input data type. So the `Spectral_data` objects are always float32/0-1. The RAYN Vision System camera for example produces uint8/0-255 multispectral images and I need to convert to float32 and rescale them to 0-1 in my pcv workflows because otherwise the results do not make sense. 

What do you think? (I'm still learning, so I might have missed something)

HaleySchuhl: I tested with and without casting to `float32` for this index, and it made no difference with our HSI format. But it sounds like your data format makes a good argument to cast `Spectral_data` during `read_data` (or even inside `_find_closest` ?) since we try to make tools flexible enough to use on different HSI types. 

@rayn-alex Would you be willing to share a small test image of your `uint8` HSI for testing? I will open an issue so we can chat more about how we're generally handling different spectral datatypes.

nfahlgren: We might need to update all the functions because I think we assumed the HSI data would be a float data type and we want to avoid integer buffer overflow. @HaleySchuhl I think you can get image data from Sam in Malia's group

rayn-alex: Of course! But maybe we should uncouple this from this PR so it can be merged, and we can create a new one for a potential additional change?

nfahlgren: Sounds reasonable!

PR #1732, opened 2025-07-15
Add color card detection squareness parameters

HaleySchuhl:
**Describe your changes**
Add optional kwargs to allow users to modify the `solidity` and `aspect_ratio` filters, implemented in `pcv.transform.detect_color_card` and within `pcv.transform.auto_correct_color`, which keeps the detected edges that are square-like. Loosening this parameters from their previously hard-coded defaults might help with color card detection in cases where the image is skewed. 

Also, adds verbose debug images of detected edges in the case where detection fails to find a color card under default settings. This aims to help debug the keyword argument settings. 

**Type of update**
Is this a:
* feature enhancement

**Associated issues**

**Additional context**
- helpful for 'nano' and other color cards that are macbeth (4x6 array of color chips) but the color card chips are more rectangular than perfectly square. 
- 
**For the reviewer**
- [x] All tests pass
- [x] Documentation tested
- [x] Code reviewed
- [ ] PR approved


joshqsumner (changes requested):
My only concern would be whether we should define aspect ratio/solidity for this context someplace in the docs. Change could be as simple as this on lines 19 and 20 of `docs/transform_detect_color_card.md` or we could do something prettier. Code wise I'd approve the change as is but could see this being helpful particularly with aspect ratio conventionally defined as width/height when what we're doing has a lower bound of 1 since we take max axis / min axis.

Edit: We could also have an if statement that takes the reciprocal of aspect_ratio if it is <1 which would help some too I think.

```
        - aspect_ratio   - Optional aspect ratio (longest side / shortest side) below which objects will get removed (default = 1.27)
        - solidity - Optional solidity (area / convex hull area) filter (default = 0.8) 
```

Issue #1582, opened 2024-08-19
Quality control check for linearity of grayscale color


kmurphy61:
**Is your feature request related to a problem? Please describe.**
I would like to check the quality of my image, and particularly my camera quality. I want to understand if my black to grey values in a color card are linear (good quality) or non-linear (representing an issue with image quality). 

**Describe the solution you'd like**
I would like to make a plot of the color chip number (1 = black, 5 = white) on the X axis, and the intensity on the Y axis (see image example below). I would like to show the expected values in a perfect scenario, the values before color correction, and values after color correction. I would like a warning threshold, so that if there is greater than 5% deviation from linear before color correction (default = 5%, but is a parameter than can be adjusted), a warning message will appear for the user to check camera settings or use an alternative image correction, such as a gamma or logv correction. I do not want the R, G, and B in the sample plot, just the grayscale chips. 

![image](https://github.com/user-attachments/assets/9aabdc8a-a233-4215-86c5-5f16c708fd20)

**Describe alternatives you've considered**
This could be part of the color card functions, or part of the quality control subpackage.

PR #1759, opened 2025-08-19
Color correction scatterplot

joshqsumner:
This is a new feature for visualizing the distance from observed to expected colors of a color card and an RGB image and optionally a color corrected image. The idea is to make it easier to assess when a camera/image is "wrong" in a bizarre way and to have a more objective gauge of how well color correction did.

This is a new feature.

This might address [1582](https://github.com/danforthcenter/plantcv/issues/1582) but I would want @kmurphy61 to weigh in on that. If the "smaller" form in 1582 would be preferred there could be an argument to this function or another function to do that.

**For the reviewer**
- [x] All tests pass
- [x] Documentation tested
- [x] Code reviewed
- [x] PR approved


NMiklave:
Yeah, this is in line with some of what I've been working on and I'd be happy to compare notes on this! Checking whether or not a color card is too worn to use requires some metric to quantify how "good" the color correction is, and I've been going about it by comparing the measured color matrix of an already-corrected image to the standard reference matrix, which can be built off of the changes made by this PR. Although, I think that the output plot of this function should flip the roles of the corrected image matrix and uncorrected image matrix. The uncorrected matrix can change quite a bit depending on lighting color temperature, but that shouldn't really matter as long as the calibrated values are consistent with what it is expected.

joshqsumner:
Thanks @NMiklave ! I'll be looking forward to seeing what you have for that, your plots a few weeks ago were very useful.

>  Although, I think that the output plot of this function should flip the roles of the corrected image matrix and uncorrected image matrix. The uncorrected matrix can change quite a bit depending on lighting color temperature, but that shouldn't really matter as long as the calibrated values are consistent with what it is expected.

I think there are two somewhat distinct use cases here. The one I was initially considering was thinking about whether there was something particularly strange in how the camera is wrong (non-linearity of error, different trends in different channels, etc) where knowing that information might help with fine tuning an imaging setup or making decisions about how the color correction should work. So the changes you mention in the uncorrected matrix are part of what I had in mind for the function to detect.
I think you're pointing out another very important setting having to do with evaluating the final quality of a color corrected image to get a handle on generalizability/accuracy.

Maybe instead of explicitly taking an uncorrected and a corrected input this could work as `color_correction_plot(std_matrix, **kwargs)` and you could pass arbitrary numbers and types of color matrices to `kwargs` then have their position/name show up in the legend? Maybe that idea is needlessly flexible and more than 2 image matrices stops being useful.

nfahlgren (changes requested):
If the change looks correct then this is good to merge

PR #1771, opened 2025-08-27
Roi filter within

### LINKED ISSUE #1651: Exclude Objects touching/overlapping ROI border
To the best of my knowledge currently the roi.filter only provides the option to either cut objects at the defined border of the roi or include objects that overlap the roi.  There is no option to exclude objects that overlap the border of the roi.  
In the images I am using, the roi borders are defined as the border of the image.  Objects that touch the roi border are not fully in the image.  

****Describe the solution you'd like****
An additional option for the roi_type that would exclude objects touching the roi border.

****Describe alternatives you've considered****

- Shrinking the roi, works in some cases, but not all.  

- Taking the image at a greater height to include more objects of interest in the image, (so the roi can be significantly shrunk without sacrificing # of objects) - however the height at which I can take an image above the plants is a constraint.

-auto_grid, however the objects are not spaced consistently across images so roi_filter works much better than auto_grid

**Additional context**
Attached a photo
If there's a way to do this that I've missed or haven't thought of (I'm pretty new to  plantcv, opencv, etc...),  would really appreciate your insight! ty!!

Image

joshqsumner:
**Describe your changes**
Adding a 'within' method to `pcv.roi.filter` based on `cutto` but filtering for only objects completely within (do not touch the border of) the ROI.

**Type of update**
Is this a feature enhancement.

**Associated issues**
Closes #1651 

**Additional context**
I don't know how common this use case really is so I might have been too specific in what I was building this towards. If there are other ideas about when/why someone would want to do this kind of filtering I'd be happy to have some discussion here about that and whether the implementation is adequate for it.

**help wanted**
Also there is a section here that could definitely use a close review and is marked with a note comment. Line 535 in `_helpers.py` I am grabbing colors from a contour placed on the original mask to see if something is masked or not. If I don't do that then I end up drawing things that were originally gaps between leaves as white sections since they are fully within the ROI. Maybe there is a better cv2 way to do that but it was not obvious to me from the docs.

**For the reviewer**
- [ ] All tests pass
- [ ] Documentation tested
- [ ] Code reviewed
- [ ] PR approved


nfahlgren:
I tested this approach with a multi-plant image where the ROI fully encompassed one plant and partially another. In `partial` mode you get both plants as expected, but in `within` mode I didn't get the fully enclosed plant and instead just got some small background noise.

I think with the hierarchical contours it gets complicated, so what if `within` mode was added to `pcv.roi.quick_filter` instead? 

If `quick_filter` was modified like this I think it does what I was expecting:

```python
def quick_filter(mask, roi, roi_type="partial"):
    """Quickly filter a binary mask using a region of interest.

    Parameters
    ----------
    mask : numpy.ndarray
        Binary mask to filter.
    roi : plantcv.plantcv.classes.Objects
        PlantCV ROI object.

    Returns
    -------
    numpy.ndarray
        Filtered binary mask.
    """
    # Increment the device counter
    params.device += 1

    # Store debug
    debug = params.debug
    params.debug = None

    # Label objects in the image from 1 to n (labeled mask)
    labels, num = label(label_image=mask, return_num=True)

    # Convert the input ROI to a binary mask (only works on single ROIs)
    roi_mask = roi2mask(img=mask, roi=roi)

    # Convert the labeled mask and ROI mask to float data types
    roi_mask = roi_mask.astype(float)
    labels = labels.astype(float)

    # Set the ROI mask value to 0.5
    roi_mask[np.where(roi_mask == 255)] = 0.5

    # Add the labeled mask and ROI mask together
    summed = roi_mask + labels

    for i in range(1, num + 1):
        # For each label, if at least one pixel of the object overlaps the ROI
        # set all the label values to the label plus 0.5
        if roi_type.upper() == "PARTIAL" and i + 0.5 in summed:
            summed[np.where(summed == i)] = i + 0.5
        # If one pixel of the object falls outside the ROI
        # set all the label values to zero
        elif roi_type.upper() == "WITHIN" and i in summed:
            summed[np.where(labels == i)] = 0

    # Objects that do not overlap the ROI will round to an integer and have
    # the same value before and after rounding.
    # Objecs that overlap the ROI will round up/down and will not have the same value
    # Where the values are equal (not overlapping)
    summed[np.where(summed == summed.round())] = 0

    # The summed image now only contains objects that overlap the ROI
    # Subtract 0.5 to remove the ROI mask
    summed = summed - 0.5

    # Round and set the data type back to uint8
    summed = summed.round().astype("uint8")

    # Make sure the mask is binary
    summed[np.where(summed > 0)] = 255

    # Print/plot debug image
    params.debug = debug
    _debug(visual=summed, filename=os.path.join(params.debug_outdir, f"{params.device}_roi_filter.png"), cmap="gray")

    return summed
```

nfahlgren:
It does work on the example I had now!

nfahlgren (approved):

PR #1803, opened 2025-09-24
Jupyter parallelization

### LINKED ISSUE #1802: Jupyter Parallelization
Running plantcv parallel from jupyter could be convenient.

Thinking that a new workflowconfig class for jupyter notebook use would let us use mostly existing backend stuff. 

Could also be a decorator in the notebook to say "run this chunk when you do parallel" and then workflowconfig could be modified to take a `.ipynb` file as the workflow (then it would nbconvert it or something i assume).

joshqsumner:
**Describe your changes**
Adding a `jupyterconfig` class to run parallel workflows from jupyter directly.
So far it just uses the existing parallel tools and the new stuff just lets you automatically convert the notebook to a script and run that script over a dataset.

**Type of update**
This is a new feature.

**Associated issues**
Closes #1802 

**Additional context**
There could be a lot of ways to implement this and I'm not sure which is best so I started with the simplest. We can always scrap it.
Depending on how this looks exactly it might be hard to test some things, the current methods mostly check if you're in a notebook before doing most of the function.

**For the reviewer**
- [ ] All tests pass
- [ ] Documentation tested
- [ ] Code reviewed
- [ ] PR approved


HaleySchuhl (approved):
Works great!

This can go in a future PR but I had minor feedback on the docs. I like the parallel_preparation.md addition. I think Why Parallelize can be generalized from what content is there now. Parallel workflows are more reproducible than what can be done with a GUI, and can we maybe add more detail about how consistent a dataset needs to be to analyze in parallel (vs when human-in-the-loop workflows are required).

PR #1878, opened 2026-02-18
Quick filter performance

nfahlgren:
**Describe your changes**
The `pcv.roi.quick_filter` method was added to improve the performance of `pcv.roi.filter`. While it is significantly faster, this PR aims to improve performance a little more. The updated approach only impacts the partial overlap method (the slowest). The core update is that the labeled mask + ROI mask is first condensed to a set of unique pixel values and these are filtered for values where a label and the ROI intersect. Then only these objects are added to the output mask. The performance of this approach will be faster when there are more objects outside the ROI because the old approach iterated over all objects.

**Type of update**

**For the reviewer**
- [ ] All tests pass
- [x] Documentation tested
- [ ] Code reviewed
- [ ] PR approved


nfahlgren:
This didn't really work, can be revisited potentially

PR #1951, opened 2026-06-24
Restores the time_lapse_video function

nfahlgren:
**Describe your changes**
Our plan was to deprecate the `pcv.visualize.time_lapse_video` function because it required QT and that created conflicts with PlantCV add-on tools. Instead we were able to refactor the function to use `imageio` instead of OpenCV and restore the functionality without QT. This PR also updates the function inputs to remove the `display` parameter, which only prints out the filepath of the video but is provided by the user. 

We might consider changing the output value. The function returns the frame size but I think it could instead not return anything, like `pcv.print_image`.

**Type of update**

**For the reviewer**
See [this page](https://docs.plantcv.org/pr_review_process/) for instructions on how to review the pull request.
- [x] All tests pass
- [x] Documentation tested
- [x] Code reviewed
- [x] PR approved

### DIFF
--- a/docs/updating.md
+++ b/docs/updating.md
@@ -1625,7 +1625,7 @@ pages for more details on the input and output variable types.
 
 * pre v4.0: NA
 * post v4.0: frame_size = **pcv.visualize.time_lapse_video**(*img_list, out_filename='./time_lapse_video.mp4', fps=29.97, display=True*)
-* post v5.0: deprecated.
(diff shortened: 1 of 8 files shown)


joshqsumner:
I changed `img_list` to `source` so that we can pass `str` inputs similar to the edits for `pixel_scatter_plot`. I was considering if there was a time that we would want to pass a list of images already in numpy form but I didn't see a situation where that would be a real workflow. I'll approve this but it might be good to bring up in dev meeting tomorrow before merging.

joshqsumner (approved):
Made some edits and it looks good, glad to keep this in v5.

PR #2001, opened 2026-08-29
Handle one, two, empty and out-of-range ENVI default bands

### LINKED ISSUE #1007: The bug from import hyperspectral image file using function readimage()
**Describe the bug**
When using function `readimage()` function to read the raw and hdr images captured from the [unispectral camera](https://www.unispectral.com/product/monarch-ii-camera/). Some bugs happen because of the size of the parameter `default_bands` 
in your source code `hyperspectral/read_data.py`, function `_maek_pseudo_rgb`,  the `default_bands` is ['713'] got from the raw file. However,  this function needs three data from the default_bands list, which is incompatible. Btw, if is reasonable to  choose the default_band in axis 2 of the array_data
```
def _make_pseudo_rgb(spectral_array):
    """Create the best pseudo-rgb image possible from a hyperspectral datacube

    Inputs:
        spectral_array = Hyperspectral data instance

    Returns:
        pseudo_rgb     = Pseudo-rgb image

    :param spectral_array: __main__.Spectral_data
    :return pseudo_rgb: numpy.ndarray
    """
    # Make shorter variable names for data from the spectral class instance object
    array_data = spectral_array.array_data
    default_bands = spectral_array.default_bands #['713']
    wl_keys = spectral_array.wavelength_dict.keys()

    if default_bands is not None:
        pseudo_rgb = cv2.merge((array_data[:, :, int(default_bands[0])],
                                array_data[:, :, int(default_bands[1])],
                                array_data[:, :, int(default_bands[2])]))
```
**To Reproduce**
Steps to reproduce the behavior (including sample data and code really help):
1. Starting with image '...'
[images1.zip](https://github.com/danforthcenter/plantcv/files/11025542/images1.zip)
2. Run function(s) 'readimage' with the `envi`mode
3. See error

**Expected behavior**
please fix it ,thanks
Or need a PR?

 - OS: ubuntu 18.04
 - Environment  virtualenv, 
 - PlantCV Version latest

**Additional context**

AbhinavMir:
#1007

### DIFF
--- a/docs/Spectral_data.md
+++ b/docs/Spectral_data.md
@@ -38,6 +38,8 @@ Attributes are accessed as spectral_data_instance.*attribute*.
 
 **filename**: The filename where the data originated from
 
+**default_bands**: The bands that are used to make the pseudo-RGB image, read from the `default bands` field of the .hdr file. Contains one band number for a grayscale image or three band numbers for an RGB image. Set to `None` if the .hdr file does not name any default bands, in which case the bands are selected from the available wavelengths
+
 **metadata**: Metadata in a dictionary. Included keys might be specific to hyperspectral or geospatial image types 
 
 ### Example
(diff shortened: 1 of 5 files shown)


joshqsumner (approved):
I do not have an example dataset where this is a problem but the author is correct about the standards in envi headers per the [docs](https://www.nv5geospatialsoftware.com/docs/ENVIHeaderFiles.html). I don't really know when we would run into a grayscale hyperspectral image in the wild but I don't see any problem with supporting it.

nfahlgren:
@all-contributors please add @AbhinavMir for code

Issue #2002, opened 2026-08-31
segment_insertion_angle: angles desync from leaves when a leaf yields 0 or 2 insertion segments


musharna:
Line references are to `plantcv/plantcv/morphology/segment_insertion_angle.py` at current `main` (dae7f5ef).

## The mechanism

The bookkeeping in `segment_insertion_angle` assumes every non-pruned leaf contributes exactly one insertion segment, but the classification loop does not guarantee that:

* Lines 89-90: for each of the two segment ends of a leaf, the end is appended to `insertion_segments` if it does not overlap a tip. Depending on tip overlap this appends **0, 1, or 2** entries per leaf.
* Line 140: `all_intersection_angles.append(intersection_angles_editing[0])` then pops one angle per non-pruned leaf, assuming the one-angle-per-leaf invariant.

Three outcomes follow, two loud and one silent (line numbers verified by running each case):

* A leaf contributes 0 insertion segments (both ends overlap tips): the editing list runs dry and line 140 raises `IndexError`.
* A leaf contributes 2 (neither end overlaps a tip, e.g. tips missing from `skel_img`): `insertion_segments` outgrows the per-leaf lists and the label loop raises `IndexError` at line 147 (`label_coord_x[i]`).
* One leaf contributes 0 and another 2, so the totals still match: no crash, but every angle between them is shifted onto the wrong leaf. Line 140 records another leaf's angle, and the colors (`rand_color[t]`, line 125) and label positions (lines 147-148) index lists that no longer run in step, so the reported angles, drawn lines, and labels silently belong to different leaves. The crashes are annoying; this silent misassignment is the case that corrupts measurements. Demonstrated concretely: a slanted stem with leaf A parallel to it (true insertion angle ~0 degrees, but absent from `skel_img` at its junction so both its ends read as tips, 0 segments) plus leaf B perpendicular to it (not in `skel_img`, so no tip overlap, 2 segments) runs without error and reports `[90.0, 90.0]`: leaf A is assigned leaf B's angle and nothing signals it.

## Evidence this was never intended: the test suite locks the crash in under another bug's name

`tests/plantcv/morphology/test_segment_insertion_angle.py::test_segment_insertion_angle_overflow` is named for the int32 slope-overflow guard, and its comment says "would usually give overflow error without extra if statement". But what it actually asserts is:

```python
with pytest.raises(IndexError):
    _ = segment_insertion_angle(...)
```

[...]

PR #2003, opened 2026-08-31
Fix segment_insertion_angle crash on near-vertical stems (int32 overflow in stem line drawing)

musharna:
**Describe your changes**

`segment_insertion_angle` crashes on a perfectly (or nearly) vertical stem. `cv2.fitLine` on a vertical stem contour returns `vx ~ 0` (about 1e-8 in practice), so extrapolating the fitted stem line to the image edges produces y-intercepts on the order of 1e9 or larger. Those exceed the 32-bit coordinate range `cv2.line` accepts, and the call raises:

```
cv2.error: OpenCV(4.11.0) :-1: error: (-5:Bad argument) in function 'line'
> Overload resolution failed:
>  - Can't parse 'pt1'. Sequence item with index 1 has a wrong type
```

An upright stem is a common case for plant images, and whether it crashes depends on image size (the extrapolated intercepts scale with image width), so the same skeleton can work on a small image and crash on a larger one.

The per-segment drawing loop a few lines below already guards against this exact failure for leaf segments (`if abs(slope) > 1000000: ... cannot be plotted`). This PR applies the equivalent protection to the stem line: if the extrapolated intercepts are not finite 32-bit values, the stem line is drawn as a vertical line at the fitted x position instead, so the debug image keeps its stem line rather than losing it. The guard checks the computed intercepts directly rather than a slope threshold because the overflow point depends on image width as well as slope.

Angle measurements are unchanged in all cases; the change only affects drawing of the debug image. No function signatures change.

Previously: vertical stem -> crash. Now: vertical stem -> same angles as before, stem line drawn vertically.

**Type of update**

* Bug fix

**Associated issues**

Related but distinct: #2002 describes a separate bookkeeping bug in the same function (this PR does not close it). The two touch different lines and do not conflict.

**Additional context**

Two regression tests are included:

* `test_segment_insertion_angle_vertical_stem`: vertical stem plus a 45 degree leaf; crashes with the `cv2.error` above on current `main`, passes with this change, and asserts the stem line is still drawn and the measured angle is ~45 degrees.
* `test_segment_insertion_angle_slanted_stem`: a slanted stem, asserting the extrapolated edge-to-edge stem line still comes from the original code path (this pins that the fallback only fires for the overflow case).

Coverage of `segment_insertion_angle.py` remains 100% (both branches of the new guard are exercised).

### DIFF
--- a/plantcv/plantcv/morphology/segment_insertion_angle.py
+++ b/plantcv/plantcv/morphology/segment_insertion_angle.py
@@ -106,9 +106,7 @@ def segment_insertion_angle(skel_img, segmented_img, leaf_objects, stem_objects,
     [vx, vy, x, y] = cv2.fitLine(combined_stem[0], cv2.DIST_L2, 0, 0.01, 0.01)
     stem_slope = -vy / vx
     stem_slope = stem_slope[0]
-    lefty = int(np.array((-x * vy / vx) + y).item())
-    righty = int(np.array(((cols - x) * vy / vx) + y).item())
-    cv2.line(labeled_img, (cols - 1, righty), (0, lefty), (150, 150, 150), 3)
+    _plot_stem_line(labeled_img, vx, vy, x, y, cols)
 
     for t, segment in enumerate(insertion_segments):
         # Find line fit to each segment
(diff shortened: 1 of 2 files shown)


musharna:
Flagging the red check, because at a glance it reads as a defect in this PR and I don't think it is.

`DeepSource: Test coverage` is failing while `DeepSource: Python` passes. The difference between them is that the coverage check needs a coverage report uploaded by a workflow run, and no workflow has run on this PR — the check list here is only the two DeepSource contexts, with no `build (3.11/3.12/3.13, ubuntu-latest)` entries at all. On #2025, which is open right now, those three build jobs are present and the coverage check is green.

So this is a missing run rather than a failing one: with no `build` job there is no coverage artifact to compare against, and the gate fails on the absence. As far as I can tell it needs a maintainer to approve the workflow run on this PR; `gh run rerun` isn't available to me here, and pushing again only re-creates the same state.

The change itself is small and I've checked it locally: `segment_insertion_angle` crashed on near-vertical stems because the extended line endpoint overflowed int32 before reaching `cv2.line`. The regression test fails with `cv2.error: ... 'line' Bad argument` when the guard is reverted and passes with it in place, rebased onto current main.

Happy to rebase, split it, or add anything you'd want to see — just let me know.

nfahlgren (approved):

nfahlgren:
@all-contributors please add @musharna for code

